PR #1, opened 2015-09-03
ASM-1973 Moving helper/util methods/classes out of asm-deployer project

asaenz:


lib/asm/network_configuration.rb
```diff
@@ -0,0 +1,369 @@
...
+          prefixes.insert(0,"NIC.Embedded.1")
+        end
+      end
+
+      if prefixes.size >= cards.size
+        prefixes.slice(0, cards.size + 1)
+      else
+        fqdds = nics.collect { |nic| nic.fqdd }
+        logger.debug("Found nic fqdd's: #{fqdds}") if logger
+        raise(ASM::UserException, t(:ASM017, "Network configuration requires %{expected_count} network cards but only %{actual_count} were found", :expected_count => cards.size, :actual_count => prefixes.size))
```
gavin-scott: I think this will be a problem without including the corresponding locale yml files. Maybe we should remove all of the translatable stuff from dell-asm-util and make this just raise a standard exception that we could catch in the calling code to display the user-facing message?

ripienaar: Responding here to a mail @asaenz sent:

So we shouldn't change the locale away from ```:en```, thats the language. If anything we can tweak the load_path to include a asm-util specific one by default and let asm-deployer override the load path as it does in ASM#init this way you always get what you want.

However, as this gem includes ```asm/translatable.rb`` the ```t()``` function will work, and the function includes a default message here to use in the absense of any specific locale or data files which means probably this will just work without issue.

Seems that way to me, we should try it but I think it will just work

gavin-scott: I forgot about the default message built from the arguments to `t()`. I think it should be ok to rely on that default when the code is used from the puppet modules and then for asm-deployer it will pick up the en.yml

asaenz: @gavin-scott @ripienaar It seems that'll work fine as long as we're strict on the version of i18n.  
I was testing locally with 0.7.0 and it wasn't working.  After testing with 0.6.9, it seems to work with a deprecation warning.
"[deprecated] I18n.enforce_available_locales will default to true in the future. If you really want to skip validation of your locale you can set I18n.enforce_available_locales = false to avoid this message."
It's easy enough to be strict in the gemspec to use 0.6.9, but should we try to set that enforce_available_locales = false anywhere, or just leave as is?

PR #3, opened 2015-09-29
ASM-4913 fix ASM::WsMan.reboot from gem refactor

gavin-scott:
ASM::WsMan.reboot was moved into dell-asm-util, but included a
static file reboot.xml as part of its arguments. The reboot.xml
was not moved into dell-asm-util causing that function to
fail. The reboot.xml file was not really needed since the
RebootJobType parameter it provided was already included as a
wsman property, so this fix just removed its use.


gavin-scott:
@asaenz please review.

asaenz:
:+1:

PR #4, opened 2015-10-22
ASM-2062 Support Post OS NIC configurations for Windows

sushilrai:
Added helper method to collect nic team information based on the networks selected on each of the partitions

### DIFF
--- a/lib/asm/network_configuration.rb
+++ b/lib/asm/network_configuration.rb
@@ -301,6 +301,9 @@ module ASM
       end
       nic_prefixes = nil
 
+      # Instance variable to track, add_nics! is invoked
+      @network_config_add_nic = true
+
       cards.each do |card|
         card.interfaces.each do |interface|
           interface.partitions.each do |partition|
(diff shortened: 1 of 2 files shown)


sushilrai:
@gavin-scott please review the changes

lib/asm/network_configuration.rb
```diff
@@ -365,5 +366,34 @@ def reset_virt_mac_addr(endpoint)
         end
       end
     end
+
+    def nic_team_info(end_point={})
```
gavin-scott: Can you add some yard-doc comments on the method and some unit tests?

lib/asm/network_configuration.rb
```diff
@@ -365,5 +366,34 @@ def reset_virt_mac_addr(endpoint)
...
+      add_nics!(end_point) unless end_point.empty?
```
gavin-scott: We should probably fail with a nice error if add_nics! wasn't already called and end_point is empty. That would require probably some instance variable to record whether add_nics! has been called.

lib/asm/network_configuration.rb
```diff
@@ -365,5 +366,34 @@ def reset_virt_mac_addr(endpoint)
...
+
+      network_info = {}
+      partitions = get_all_partitions
+      unless partitions.empty?
```
gavin-scott: This isn't needed, the loop on next line will just do nothing it is `empty?`

lib/asm/network_configuration.rb
```diff
@@ -365,5 +366,34 @@ def reset_virt_mac_addr(endpoint)
...
+            network_info[networks] ||= []
+            network_info[networks].push(partition.mac_address)
+          end
+        end
+      end
+      @teams = []
+      network_info.each do |network,macs|
+        @teams.push({:networks => network , :mac_addresses => macs})
+      end
+      @teams
```
gavin-scott: Since you're saving this it would probably make sense to memoize this. So this whole method would become something like:

    def teams
      @teams ||= begin
        # your logic...
      end
    end

lib/asm/network_configuration.rb
```diff
@@ -365,5 +366,34 @@ def reset_virt_mac_addr(endpoint)
...
+          networks = begin
+            (partition.networkObjects || []).collect {
+                |network| network.staticNetworkConfiguration.delete('ipRange') if !network.staticNetworkConfiguration.nil?
+            }
+            partition.networkObjects.reject { |network| network.type == 'PXE' }
+          end.flatten.uniq.compact
+
+          # Need to find partitions which has same set of networks, for team
+          if networks && !networks.empty?
+            network_info[networks] ||= []
```
gavin-scott: The use of the networks array of hashes as a key here makes me nervous. What if the networks on different partitions are in a different order, does that break this?

lib/asm/network_configuration.rb
sushilrai: I removed this intentionally, as we can add networks dynamically and we have to update the NIC Team information

lib/asm/network_configuration.rb
sushilrai: I tried selecting two networks on separate partitions in different sequence but in the network configuration object I got them in the same sequence.

lib/asm/network_configuration.rb
gavin-scott: What's the case where add networks dynamically?

sushilrai: Network scale-up.

lib/asm/network_configuration.rb
gavin-scott: It would probably be good to sort the networks for the key anyway. If they ever get passed in a different order it will lead to subtle bugs. I think you could just do:

    networks = networks.sort_by { |k| k["id"] }

to sort them by the network id.

lib/asm/network_configuration.rb
gavin-scott: But isn't that included in the `network_configuration` data passed in during the scale-up? I'm not aware of a case where we change the `network_configuration` data programatically during a deployment (scale-up, retry etc)

sushilrai:
@gavin-scott please take a look into the changes.

PR #5, opened 2015-10-24
remove ipRange from networkObjects in network_configuration

gavin-scott:
The ipRange field contains a list of ip ranges associated with the
given network. It is sometimes passed in different orders, which can
cause networkObjects that are ostensibly same (the same network id and
same static IP if it is a static network) to appear different. That
causes difficulty when doing things like trying to determine the set
of networks that are being used on a port.


gavin-scott:
@sushilrai or @ripienaar please review.

PR #7, opened 2015-10-30
Added support for attach/boot network iso

asaenz:
DO NOT MERGE: Still testing.

@gavin-scott I don't know if we ever plan on using this for any other kind of ISO other than the small one, so the hour long timeout I have right now might be WAY overkill.
Just review and let me know if there's anything you want to change here, including that.


gavin-scott:
Merging this as is, I'll work on updating it as needed.

PR #10, opened 2015-11-26
ASM-4621 choose best NICs to match NetworkConfiguration

gavin-scott:
Previously server NICs were assigned to NetworkConfiguration
interfaces solely based on NIC order. I.e. if the user chose a 4x10Gb
NIC and and 2x10Gb NIC it was assumed that the server should have a
4x10Gb NIC before a 2x10Gb NIC according to slot order. New versions
of ASM allow more flexible NIC matching, i.e. in the previous example
a server that had a 2x10Gb NIC before a 4x10Gb NIC would also be
matched.

This commmit updates ASM::NetworkConfiguration to match NICs in a
similar fashion based on NIC capabilities. For each NIC requested in
the NetworkConfiguration a server NIC with matching capabilities (link
speed and number of NPAR partitions) will be found.

Link speed can be determined programatically via WS-Man in most
cases. However, if not available this code will fall back to
hard-coded information based on certain known NIC models such as the
Broadcom / QLogic 578xx series or the Intel X520. Number of NPAR
partitions can only be determined based on hard-coded NIC model
information for the Broadcom 578xx series; all other NICs are treated
as not supporting NPAR.

### DIFF
--- a/lib/asm/network_configuration.rb
+++ b/lib/asm/network_configuration.rb
@@ -1,9 +1,10 @@
-require 'asm/errors'
-require 'asm/wsman'
-require 'asm/translatable'
-require 'hashie'
-require 'asm/network_configuration/nic_info'
-require 'asm/network_configuration/nic_type'
+require "asm/errors"
+require "asm/wsman"
+require "asm/translatable"
+require "hashie"
+require "asm/network_configuration/nic_view"
+require "asm/network_configuration/nic_type"
+require "asm/network_configuration/nic_info"
 
 module ASM
   # The NetworkConfiguration class is a wrapper class to make it easier to work
@@ -63,7 +64,7 @@ module ASM
     def get_wsman_nic_info(endpoint)
       fqdd_to_mac = ASM::WsMan.get_mac_addresses(endpoint, logger)
       fqdd_to_mac.keys.map do |fqdd|
-        nic = NicInfo.new(fqdd, logger)
+        nic = NicView.new(fqdd, logger)
         nic.mac_address = fqdd_to_mac[fqdd]
         nic
       end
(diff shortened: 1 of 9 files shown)


lib/asm/network_configuration/nic_info.rb
```diff
@@ -1,51 +1,172 @@
+require "asm/network_configuration/nic_port"
```
gavin-scott: Note for review: this should be treated as a new file for review. I renamed the old NicInfo class to NicView and this is effectively a new class. NicInfo now represents a single physical server NIC capabilities whereas NicView represents the WS-Man data retrieved for a single partition on a NIC.

gavin-scott:
@sushilrai or @ripienaar can you review? I know this is a lot of code, but it's needed in order to be able to match NICs more intelligently. If you want me to review some of it with you on a call let met know.

ripienaar:
Not sure I am going to have the time to devote to this one it deserves in the next 2 days

gavin-scott:
@ripienaar are you ok if I get @asaenz to review this? He may be a bit more familiar with the `network_configuration`

ripienaar:
@gavin-scott yeah, me reviewing this one will be a struggle to really review it for suitability

gavin-scott:
@asaenz do you mind reviewing this? I can sit down with you after planning this morning and go over the details of what it's trying to accomplish -- classify NICs according to their type and capabilities (2x10Gb the supports NPAR, etc) and use that info to more intelligently match the NICs in the network_configuration data.

lib/asm/network_configuration/nic_info.rb
```diff
@@ -1,51 +1,172 @@
...
+        !!(nic_status =~ /disabled/i)
+      end
+
+      # If all ports have the same link speed
+      #
+      # @param ports [Array<NicPort>] the ports
+      # @param link_speed [String] the link speed
+      # @return [Boolean] if all ports have the same link speed
+      def all_ports?(ports, link_speed)
+        ports.map { |p| p.link_speed }.uniq == [ link_speed ]
```
asaenz: I don't really like this.  It just looks strange to me.  Could also do something like:
>ports.all? { |p| p.link_speed == link_speed}

To me it's just a little more clear on how it's working, but that's pretty nit picky.

gavin-scott: Yep, that's nicer. Will update.

lib/asm/network_configuration/nic_info.rb
```diff
@@ -1,51 +1,172 @@
...
+
+      # The number of NPAR partitions for each 10Gb port on the NIC
+      #
+      # NICs that do not support NPAR will return 1
+      #
+      # @raise [StandardError] if different 10Gb ports report different partitioning capabilities
+      def n_partitions
+        ports_10gb = ports.find_all { |port| port.link_speed == "10 Gbps" }
+        ns = ports_10gb.map { |port| port.n_partitions }.uniq
+        return 0 if ns.empty?
```
asaenz: Is this the case when the NIC doesn't support NPAR?  The comment above mentions this returns 1 in that case.

gavin-scott: Doh, nice one! Yes, should return 1. Will update.

gavin-scott:
@asaenz I've updated based on your comments and added more tests for NicInfo. Do you want to give it another look?

lib/asm/network_configuration.rb
```diff
@@ -242,38 +243,12 @@ def compare_cards(nic1, nic2)
...
-      if prefixes.size >= cards.size
-        prefixes.slice(0, cards.size + 1)
-      else
-        fqdds = nics.collect { |nic| nic.fqdd }
-        logger.debug("Found nic fqdd's: #{fqdds}") if logger
-        raise(ASM::UserException, t(:ASM017, "Network configuration requires %{expected_count} network cards but only %{actual_count} were found", :expected_count => cards.size, :actual_count => prefixes.size))
-      end
+    def n_partitions(card)
+      ns = card.interfaces.map { |i| i.partitions.size }.uniq
+      return 0 if ns.empty?
```
asaenz: Should this be returning 0?  The other n_partitions method was returning 1, but it's hard to tell the different use cases for each.

gavin-scott: The question of how many partitions zero interfaces have is an interesting one. But I think you're right that for consistency it makes sense to default to 1.

asaenz:
I think this probably looks fine, but it's honestly hard to tell at a glance without fully testing myself. If bugs come through, it's no fault of mine.


:+1:

PR #11, opened 2015-12-10
Intel NIC having incorrect Current MAC Address

sushilrai:
In case NIC is having default current MAC Address then pick permanent MAC Address instead. This scenario is observed with Intel NICs

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -216,11 +216,21 @@ module ASM
     def self.get_mac_addresses(endpoint, logger = nil)
       bios_info = get_bios_enumeration(endpoint,logger)
       ret = get_nic_view(endpoint, logger).inject({}) do |result, element|
-        result[element['FQDD']] = element['CurrentMACAddress'] if is_usable_nic?(element,bios_info)
+        result[element['FQDD']] = select_mac_address(element) if is_usable_nic?(element, bios_info)
         result
       end
       logger.debug("********* MAC Address List is #{ret.inspect} **************") if logger
       ret
+    end
+
+    def self.select_mac_address(element)
+      if element['CurrentMACAddress'] != '00:00:00:00:00:00'
+        element['CurrentMACAddress']
+      elsif !element['PermanentMACAddress'].nil?
+        element['PermanentMACAddress']
+      else
+        nil
+      end
     end
 
     def self.get_permanent_mac_addresses(endpoint, logger = nil)
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -226,7 +226,7 @@ module ASM
     def self.select_mac_address(element)
       if element['CurrentMACAddress'] != '00:00:00:00:00:00'
         element['CurrentMACAddress']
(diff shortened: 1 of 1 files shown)


lib/asm/wsman.rb
```diff
@@ -216,8 +216,14 @@ def self.bios_display_name(fqdd)
...
       bios_info = get_bios_enumeration(endpoint,logger)
       ret = get_nic_view(endpoint, logger).inject({}) do |result, element|
-        result[element['FQDD']] = element['CurrentMACAddress'] if is_usable_nic?(element,bios_info)
-        result
+        if is_usable_nic?(element,bios_info)
+          if element['CurrentMACAddress'] != '00:00:00:00:00:00'
+            result[element['FQDD']] = element['CurrentMACAddress']
+          else
+            result[element['FQDD']] = element['PermanentMACAddress']
+          end
```
gavin-scott: I was originally thinking we should check that the mac is not an empty string also or otherwise mal-formed, although we haven't seen that happen. But you could create a method like `valid_mac?` to encapsulate that and change this to look like:

    mac = element["CurrentMACAddress"]
    mac = element["PermanentMACAddress"] unless valid_mac?(mac)

gavin-scott:
Travis CI is failing, looks like some tests will need to be updated. Would be good to specifically add a test for the `00:00:00:00:00` case as well.

sushilrai:
@gavin-scott please review

lib/asm/wsman.rb
```diff
@@ -216,13 +216,24 @@ def self.bios_display_name(fqdd)
     def self.get_mac_addresses(endpoint, logger = nil)
       bios_info = get_bios_enumeration(endpoint,logger)
       ret = get_nic_view(endpoint, logger).inject({}) do |result, element|
-        result[element['FQDD']] = element['CurrentMACAddress'] if is_usable_nic?(element,bios_info)
+        select_mac_address(result, element) if is_usable_nic?(element,bios_info)
```
gavin-scott: I'd really prefer to have select_mac_address return the mac address rather than modify the result array. Generally it's best (creates fewer confusing bugs) to avoid mutating method arguments where possible. Please update `select_mac_address` to just return the mac address so above line becomes:

    result[element['FQDD']] = select_mac_address(element) if is_usable_nic?(element, bios_info)

sushilrai:
@gavin-scott Please review

lib/asm/wsman.rb
```diff
@@ -216,13 +216,23 @@ def self.bios_display_name(fqdd)
...
         result
       end
       logger.debug("********* MAC Address List is #{ret.inspect} **************") if logger
       ret
     end
 
+    def self.select_mac_address(element)
+      if element['CurrentMACAddress'] != '00:00:00:00:00:00'
+        element['CurrentMACAddress']
+      elsif !element['PermanentMACAddress'].nil?
```
ripienaar: ```elsif element["PermanentMACAddress"]```

sushilrai:
@ripienaar please review

ripienaar:
:+1:

PR #12, opened 2015-12-10
ASM-5382 Added a new error for switch-server connectivity failures

asaenz:


lib/asm/errors.rb
```diff
@@ -9,4 +9,14 @@ class NagiosException      < Error; end
 
   # A UserException message can be displayed directly to the user
   class UserException        < Error; end
+
+  class UnconnectedServerException < Error
```
gavin-scott: Would it make more sense to just throw this in ASM::Service::SwitchCollection where you are now throwing it in https://github.com/dell-asm/asm-deployer/pull/1225 ? I don't think it make sense to add asm-deployer specific errors to dell-asm-util.

asaenz: Yeah,I wasn't sure about that.  It is ASM specific, but I also thought it might make sense to keep the exceptions in one place.  Maybe we need a new private_errors file or something?

asaenz:
This exception was just kept inside asm-deployer, since it's asm-deployer specific

PR #13, opened 2015-12-22
ASM-5219 new ISO commands and better response parsing

gavin-scott:
This PR is more preparation for supporting O/S installs via virtual media mount of ipxe.iso

- add support for DeploymentService ConnectNetworkISOImage, GetAttachStatus and GetNetworkISOConnectionInfo methods.

- add poll_for_lc_ready which will automatically disconnect ISO images. Those connections would block the LC from ever becoming ready. This is an improvement on wait_for_lc_ready.

- add parse method for parsing ws-man responses into ruby-like hashes. Also detects and parses ws-man fault responses.

- add snake_case and camel_case methods to make it easier to accept ws-man invoke command parameters as ruby-style hashes.

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -2,16 +2,66 @@ require 'pathname'
 require 'asm/util'
 require 'rexml/document'
 require 'hashie'
+require 'nokogiri'
+require 'logger'
 
 module ASM
   module WsMan
 
     class Error < StandardError; end
 
-    DEPLOYMENT_SERVICE_SCHEMA = 'http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_OSDeploymentService?SystemCreationClassName="DCIM_ComputerSystem",CreationClassName="DCIM_OSDeploymentService",SystemName="DCIM:ComputerSystem",Name="DCIM:OSDeploymentService"'
+    # Convert a wsman response hash into a human-readable string.
+    #
+    # @example
+    #     response = {:message => "Could not connect to share", :code => "XXX", :return_value => "2"}
+    #     ASM::WsMan.response_string(response) #=> "Could not connect to share [code: XXX, return_value: 2]"
+    #
+    # @api private
+    # @param response [Hash] ws-man response as a Hash, i.e. after calling {#parse} on the response.
+    # @return [String]
+    def self.response_string(response)
+      copy = response.dup
+      message = copy.delete(:message)
+      message = copy.delete(:reason) unless message
+      message = copy.delete(:job_status) unless message
+      ret = copy.keys.map { |k| "%s: %s" % [k, copy[k]]}.join(", ")
+      ret = "%s [%s]" % [message, ret] if message
+      ret
+    end
 
-    # Wrapper for the wsman client. endpoint should be a hash of
-    # :host, :user, :password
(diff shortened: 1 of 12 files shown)


gavin-scott:
@ripienaar would you mind reviewing sometime this week? I know you're not very familiar with the domain but I'd like to get this in as the base for more ws-man command flows I'm going to have to implement for the ipxe ISO boot stuff.

ripienaar:
@gavin-scott will do 'morrow :)

lib/asm/wsman.rb
```diff
@@ -357,66 +462,404 @@ def self.get_fcoe_wwpn(endpoint, logger = nil)
...
+    # @example busy response
+    #   {:lcstatus=>"5", :message=>"Lifecycle Controller Remote Services is not ready.",
+    #    :message_id=>"LC060", :rtstatus=>"1", :return_value=>"0", :server_status=>"1", :status=>"1"}
+    #
+    # An lcstatus of "0" indicates that the LC is ready to accept new jobs.
+    #
+    # @param endpoint [Hash] the server connection details. See {invoke} endpoint hash.
+    # @option options [Hash] :logger
+    # @return [Hash]
+    def self.get_lc_status (endpoint, options = {})
```
ripienaar: odd spacing here

lib/asm/wsman.rb
```diff
@@ -357,66 +462,404 @@ def self.get_fcoe_wwpn(endpoint, logger = nil)
...
-        rescue ASM::WsMan::Error
-          logger.debug("Invalid response...job may not have been initialized yet.  Waiting...") if logger
+    # Connect a network ISO as a virtual CD-ROM
+    #
+    # The normal server boot order will be ignored after this call has been made.
+    # The server will only boot into the network ISO until {disconnect_network_iso_image}
+    # is called. The LC controller will be locked while the server is in this
+    # state and no other LC jobs can be run.
+    #
+    # @param endpoint [Hash] the server connection details. See {invoke} endpoint hash.
```
ripienaar: in cases where all the options are the same you can just do ```@param (see #invoke)``` and it will copy them - not sure how that handles individual parameters though, suspect it doesn't which is a shame

tasks/rubocop.rake
```diff
@@ -0,0 +1,4 @@
+desc "Run rubocop style and lint checks"
+task :rubocop do
```
ripienaar: should we add this to the default check so travis will run it or still some other things to fix?

ripienaar:
this looks really good apart from minor niggles in comments

lib/asm/wsman.rb
gavin-scott: Yeah I spent awhile trying to get that to work but wasn't able to. When you do `@param (see #invoke)` it copies in the invoke parameter documentation exactly, including parameters that don't exist for this method. I ended up going with this style of just referencing `invoke`.

tasks/rubocop.rake
gavin-scott: Still a ton to fix. I figured I would do the clean-ups in a follow-up PR and enable it by default.

gavin-scott:
Thanks for looking at it, updated the `get_lc_status` spacing.

ripienaar:
Looks good to merge then I think

PR #14, opened 2015-12-30
maint: correct rubocop style offenses

gavin-scott:
Correct all rubocop offenses and enable it to run with the default rake task.


ripienaar:
wow, hope the bulk of these were done with its auto fixup :)

gavin-scott:
Hah, a fair number were, but I spent more time on it that I should have. One of the files was throwing the auto-correct into an infinite loop!

PR #15, opened 2016-01-08
ASM-5498 fix deployments embedded Intel X520 NICs

gavin-scott:
Old switch configuration code paths use ASM::WsMan.get_mac_addresses
to find all usable 10Gb mac addresses. It called
ASM::WsMan.is_usable_nic? which had some simple hard-coded rules about
which macs were usable. One of those was to only allow Embedded NICs
if they were Broadcom 57810s. Newer code paths use
ASM::WsMan::NetworkConfiguration which has enhanced rules for
determining whether NICs are usable 10Gb NICs such as inspecting the
NIC capability data from WSMan. Since the old switch configuration
paths haven't been completely removed they were causing failures on
embedded intel X520 deployments. This fixes that up by changing
ASM::WsMan.get_mac_addresses to reuse the NetworkConfiguration paths
and removing the old ASM::WsMan.is_usable_nic? method.

Also made these changes which are refactors in the
NetworkConfiguration code that were necessary to allow it to be reused
by ASM::WsMan.get_mac_addresses :

- Store all partition NICViews in the NICPort. Previously only
  partition 1 info was stored.

- Make NICView a bit easier to work with by responding to the []
  operator with the original hash data.


gavin-scott:
@asaenz can you review?

gavin-scott:
@asaenz can you give this a once-over? I updated the PR this morning to include fixes for a couple comments that @ripienaar had in a separate PR that included this commit: https://github.com/dell-asm/dell-asm-util/pull/17/files#r49299228 and https://github.com/dell-asm/dell-asm-util/pull/17/files#r49299358

asaenz:
:+1:

PR #16, opened 2016-01-08
ASM-4621 use NICs in UI-specified order if available

gavin-scott:
Allow network configuration data to specify the exact NIC to use for
an interface rather than always picking it based on type and number of
partitions.

Additionally add the chosen NIC fqdd to the interface so
that the resulting data is consistent whether the network
configuration specified the NIC or it was chosen by add_nics!

### DIFF
--- a/lib/asm/network_configuration.rb
+++ b/lib/asm/network_configuration.rb
@@ -265,7 +265,18 @@ module ASM
 
       missing = []
       cards.each do |card|
-        index = nics.find_index { |n| !n.disabled? && n.nic_type == card.nictype.nictype && n.n_partitions >= n_partitions(card) }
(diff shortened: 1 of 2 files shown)


gavin-scott:
@asaenz please review. Should be merged after #15 

DO NOT MERGE for time being. Needs end-to-end testing and there appears to be a bug in the NICs that the UI picks in certain cases. I will file separate jira.

gavin-scott:
Tested and ready but DO NOT MERGE until ASM-5507 has been addressed. Would break C6320 deployments.

asaenz:
:+1:

gavin-scott:
Fix for ASM-5507 was merged to asm-core. Merging this as well.

PR #17, opened 2016-01-10
ASM-5219 WsMan refactor and boot from RFS ISO implementation

gavin-scott:
The existing methods for booting from a remote ISO image as a virtual
CD were simple and effective but had disadvantages that prevent using
them to perform an iPXE O/S installation via virtual media. They
prevent other jobs from running on the iDrac and (more critically)
temporarily replace the normal boot order with one that only boots the
virtual CD. Because of the latter problem an iPXE O/S install cannot
chain-load to boot off of the hard drive.

This PR implements support for the various RFS (remote file
system) ISO methods in the OS Deployment Profile. Those create virtual
CD device that is treated like other devices in the boot
order. However actually using that device then requires implementing
support for a bunch of iDrac workflows such as setting the boot
order. Those workflows are included here.

Additionally a large number of refactors to the ASM::WsMan module
are included:

- Make WsMan a class that encapsulates the endpoint and logger so
  those don't have to be passed to all methods.

- Separate out CLI invocation and request/response parsing into
  separate classes.

- Add WsMan::Client #invoke, #get and #enumerate helper methods
  to make it much simpler to implement support for new ws-man
  commands of those types.


gavin-scott:
This includes the commits from #15 and #16 ; those should be merged first.

lib/asm/network_configuration/nic_view.rb
```diff
@@ -5,24 +5,26 @@ class NetworkConfiguration
...
       def create_with_partition(partition)
-        NicView.new(@mash.fqdd.gsub(/[-]\d+$/, "-#{partition}"))
+        NicView.new(fqdd.gsub(/[-]\d+$/, "-#{partition}"))
       end
 
-      # Forward methods we don't define directly to the mash
+      # Forward [] method directly to raw NICView hash data
       def method_missing(sym, *args, &block)
-        @mash.send(sym, *args, &block)
+        return @raw_nic_view.send(sym, *args, &block) if sym == :[]
```
ripienaar: why not just ```def []``` and remove the method_missing? I don't think it does anything anymore - in the previous code it caught everything and called the mash, now its just the one so implementing just that might be clearer?

lib/asm/network_configuration/nic_view.rb
```diff
@@ -30,32 +32,45 @@ def card_to_fabric(card)
...
-        ret.fqdd = fqdd
-        (_, ret.type, port_info) = ret.fqdd.split(".")
-        (ret.card, ret.port, ret.partition_no) = port_info.split("-")
-        ret.partition_no = "1" if ret.partition_no.nil?
-        if ret.card =~ /([0-9])([A-Z])/
-          orig_card = ret.card
-          ret.card = $1
-          ret.fabric = $2
+        @fqdd = fqdd
+        (_, @type, port_info) = @fqdd.split(".")
```
ripienaar: might be worth checking the format of the FQDD before splitting it and assuming it's in the right format?

lib/asm/wsman.rb
```diff
@@ -319,33 +274,39 @@ def self.bios_display_name(fqdd)
       display_name
     end
 
-    # Return all the server MAC Address along with the interface location
-    # in a hash format
+    # Return all 10Gb, enabled current server MAC Address along with the interface
+    # location in a hash format.
+    #
+    # @deprecated Use {ASM::NetworkConfiguration::NicInfo} instead to find NIC capabilities
```
ripienaar: I wonder if we should maybe log some info/warning when deprecated methods get used?

lib/asm/wsman.rb
```diff
@@ -823,40 +1013,175 @@ def self.boot_to_network_iso(endpoint, source_address, logger=nil, image_name="m
...
+      unless boot_mode[:current_value] == "Bios"
+        # Set back to bios boot mode
+        logger.info("Current boot mode on %s is %s, resetting to Bios BootMode" %
+                        [host, boot_mode[:current_value]])
+        set_bios_attributes(:target => boot_mode[:fqdd], :attribute_name => "BootMode",
+                            :attribute_value => "Bios")
+      end
+
+      target = find_boot_device(boot_device)
+      unless target
```
ripienaar: I quite like merging these 2 lines ie.

```unless target = find_boot_device(...)```

not sure if that's just me  - I think the style guide doesnt support that as a good idea but it seems really nice to me, what do you think?

ripienaar:
this all looks really good to me apart from those tiny comments

ripienaar:
Of course might be good for someone else to look as well who knows the ISO stuff better? I think @sushilrai knew this?

lib/asm/wsman.rb
gavin-scott: I have mixed feeling about that. I like the concision of that style and used to use it but quit since most automated tools complain about it. Here's what the style guide has to say:

https://github.com/bbatsov/ruby-style-guide/blob/master/README.md#user-content-safe-assignment-in-condition

How do you feel about adding parentheses like they suggest:

    unless (target = find_boot_device(...))

lib/asm/network_configuration/nic_view.rb
gavin-scott: That's a good idea, I'll update.

lib/asm/network_configuration/nic_view.rb
gavin-scott: Agree, will update.

lib/asm/wsman.rb
gavin-scott: That's not a bad idea. I'll throw that in.

lib/asm/wsman.rb
ripienaar: yeah, that's good too - I havnt seen MRI complain about it but ()'s are fine

gavin-scott: K, I'll go that route.

gavin-scott: Hmm, I hadn't noticed MRI complain about it either but it looks like it does:

    [gavin@localhost dell-asm-util]$ irb
    1.9.3-p551 :001 > if a  = "a"; puts "yep: %s" % a; end
    (irb):1: warning: found = in conditional, should be ==
    yep: a
     => nil

ripienaar: wth, ok :) probably best avoided then, thats annoying

lib/asm/network_configuration/nic_view.rb
gavin-scott: This comment and the one above are really part of PR #15  I've updated that with these changes, it should be merged before this one.

lib/asm/wsman.rb
gavin-scott: On second thought, I'm slightly nervous about breaking things if I add these logger.warns. A couple of reasons that's a bit more complicated than it seems:

- logger is optional so might be nil
- puppet module code passes `Puppet` as the logger which doesn't have a `.warn` method. It's called `.warning` there so I'd have to make sure we call the right one somehow.

Would rather leave this as is and address in future by transitioning clients to use the new methods.

ripienaar: That sounds fine yeah

gavin-scott:
This PR is no longer dependent on any of the others, should be OK to merge. @ripienaar has code reviewed but asked if anyone with more domain knowledge wanted to review. 

@asaenz or @sushilrai do you want to have a look? These are probably the interesting bits in terms of actual ws-man workflow:

Boot RFS ISO: https://github.com/gavin-scott/dell-asm-util/blob/c16ab7cdc9c382433f6584e8a55d4e2eec36665e/lib/asm/wsman.rb#L1166

Set boot order: https://github.com/gavin-scott/dell-asm-util/blob/c16ab7cdc9c382433f6584e8a55d4e2eec36665e/lib/asm/wsman.rb#L1105

gavin-scott:
I'm going to merge based on lack of further comments and @ripienaar 's original review.

PR #18, opened 2016-01-13
ASM-5219 updates to support iPXE ISO boot

gavin-scott:
- Add NicView data to the NetworkConfiguration partition data so that it is possible to determine what type of NIC is being used for e.g. the PXE partitions.

- Set boot order based on instance id rather than name. That makes it easy to set a specific NIC in the boot order based on its FQDD.


gavin-scott:
@asaenz can you review? This should be good to go, my end to end tests are passing.

asaenz:
:+1:

PR #19, opened 2016-01-15
maint: warn when the proxy is set

ripienaar:


ripienaar:
@gavin-scott not sure if this is desirable, but it would have saved me a ton of pain today

lib/asm/wsman/client.rb
```diff
@@ -13,6 +13,15 @@ def initialize(endpoint, options={})
...
         @endpoint = endpoint
         @logger = augment_logger(options.delete(:logger) || Logger.new(nil))
+
+        proxy_warn
+      end
+
+      # @api private
+      def proxy_warn
+        if ENV.include?("http_proxy") || ENV.include?("https_proxy")
+          logger.warn("wsman invocations will use the proxy set in http_proxy or https_proxy")
```
ripienaar: oh, puppet doesnt have warn, let me make it info instead? else I can also augment puppet with warn below

gavin-scott: Yeah, if you add it to the augment, that would work. Other than that I think this seems ok.

ripienaar: done in a follow up commit

PR #20, opened 2016-01-15
maint: Update to latest rubocop and fix offenses

gavin-scott:


lib/asm/util.rb
```diff
@@ -331,11 +331,11 @@ def self.load_file(filename, base_dir=INSTALLER_OPTS_DIR)
...
-        if File.extname(filename) == ".yaml"
-          result = YAML.load_file(filepath)
-        else
-          result = File.read(filepath)
-        end
+        result = if File.extname(filename) == ".yaml"
+                   YAML.load_file(filepath)
+                 else
+                   File.read(filepath)
+                 end
```
gavin-scott: Hmm, I thought this was the style you don't like @ripienaar -- `rubocop -a` did this. I pulled in the asm-deployer .rubocop.yml in this PR.

ripienaar: ah yeah, I'm not keen on these at all :( disabled them on asm-deployer, we should probably somehow harmonize the .rubocop.yamls across projects

gavin-scott: Odd thing was the .rubocop.yml here is the same as current asm-deployer (I copied it from there)

gavin-scott: Hmmm, sort of seems like my .rubocop.yml isn't being used. It should disable the `casecmp` stuff also and there are changes here that change `.downcase` comparisons to use `casecmp`...

ripienaar: weird, I just fix them by hand generally, find it weirdly relaxing heh

gavin-scott:
I've fixed up the changes that `rubocop -a` made that I didn't like, this should be good to go.

ripienaar:
:+1:

PR #21, opened 2016-01-16
Implement IPMI client similar to WsMan client

gavin-scott:
Implement common operations power_state, reboot, power_on and power_off using the ipmitool client.


ripienaar:
nice

PR #22, opened 2016-01-16
Various fixes and enhancements

gavin-scott:
- Pass :reboot_job_type through from boot_rfs_iso_image to set_boot_order

- Remove unused NetworkConfiguration#is_blade?

- Implement WsMan #power_on and #power_off as instance methods

- Return :on or :off from #power_state rather than a string

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1179,7 +1179,7 @@ module ASM
         find_boot_device(:virtual_cd) || raise(RetryException)
       end
 
-      set_boot_order(:virtual_cd)
+      set_boot_order(:virtual_cd, options)
 
     rescue Timeout::Error
       raise(Error, "Timed out waiting for virtual CD to become available on %s" % host)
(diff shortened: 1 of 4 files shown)


gavin-scott:
@ripienaar with this and https://github.com/dell-asm/dell-asm-util/pull/21 ASM::WsMan and ASM::Ipmi should consistently implement power_state, power_on, power_off and reboot.

lib/asm/wsman.rb
```diff
@@ -168,13 +169,16 @@ def boot_source_settings
       client.enumerate("http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_BootSourceSetting?__cimnamespace=root/dcim")
     end
 
-    # Get power state information
+    # Get server power state
     #
-    # @return [String] The value will be "2" if the server is on and "13" if it is off.
+    # @return [Symbol] :on or :off
```
ripienaar: you could also do ```@return [:on,:off]``` :)

ripienaar:
love it

lib/asm/wsman.rb
gavin-scott: Ah, thanks, good to know.

PR #24, opened 2016-01-26
ASM-5141 capture NIC info in NetworkConfiguration#to_hash

gavin-scott:
Previously it was just returning the original network_configuration
data that was passed in from the UI. However that ignores the physical
NIC information that is captured when #add_nics! is called such as
FQDD and mac address. We would like to capture that data so that the
network configuration can be passed to razor ERB templates and used to
find out which NICs should be used for PXE boot.

### DIFF
--- a/lib/asm/network_configuration.rb
+++ b/lib/asm/network_configuration.rb
@@ -354,8 +354,43 @@ module ASM
       end
     end
 
+    # Returns the network configuration data in hash form
+    #
+    # The returned value is roughly analogous to the hash that the NetworkConfiguration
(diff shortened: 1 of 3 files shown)


gavin-scott:
I'm not wild about this code but not sure offhand what would be better. We need to be able to use the `to_hash` data to pass as yaml data for puppet configuration runs (in which case complex objects like NicPort and NicView will break deserialization on the puppet side) and we also want to be able to have the phyical NIC information available when we pass the data through to razor ERB templates.

ripienaar:
yeah cant think of a immediate good improvement either

PR #25, opened 2016-01-26
Add wsman_shell executable to gem

gavin-scott:
Executable allows for simple execution of wsman commands.


gavin-scott:
I found this useful. It does add a dependency on pry though...

ripienaar:
looks handy yeah - style fails though, need to freeze constants

gavin-scott:
Updated for style checks.

PR #26, opened 2016-01-26
ASM-5660 strip out NicInfo in NetworkConfiguration#to_hash

gavin-scott:
Complex types were mistakenly being returned from #to_hash after
recent changes resulting in broken puppet yaml data in some cases.


gavin-scott:
@asaenz please review. This fixes the issue with weird network config data we saw.

asaenz:
:+1:

PR #27, opened 2016-02-01
ASM-3711 esxcli commands will now time out properly

asaenz:


lib/asm/util.rb
```diff
@@ -133,9 +133,7 @@ def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=60
         logger.debug("Executing esxcli #{tmp.join(' ')}")
       end
 
-      result = Timeout.timeout(time_out) do
-        ASM::Util.run_command_with_args("env", *args)
-      end
+      result = ASM::Util.run_command_with_args("timeout", *args)
```
ripienaar: This seems a tad fragile. Requiring that the first argument should be a number else things go bad? We'd def want this documented and tested for in the method that the first one is indeed a number of we do go this way

asaenz: Well, the first argument is required to a number, yes...but not from calling the esxcli method. I'm not totally sure what the concern is here? The method is still called without passing in an integer (unless you want a shorter timeout, an example in my other PR https://github.com/dell-asm/asm-deployer/pull/1330/files#diff-c77806842f5fda02ee42d07d36fc774aR487)

ripienaar: You're right I think it's fine as is actually. That's what I get for reviewing code on my phone :)

asaenz: Yay! I like it when I'm right :confetti_ball:

PR #31, opened 2016-02-11
ASM-5589 Allow using URI string to specify ISO boot location

gavin-scott:
URI strings like nfs://hostname/path/to/boot.iso or
smb://hostname/path/to/boot.iso can now be used in the various ISO
connection methods rather than the rather cumbersome individual ws-man
parameters.

Also testing uncovered that ConnectRFSISOImage uses an inconsistent
parameter for username -- :username instead of :user_name. Update that
method to use :username


gavin-scott:
This PR is mainly convenience to make it easier to specify the ISO URLs. But also fixes a bug that prevented CIFS shares from being used with ConnectRFSISOImage -- the method expects `username` instead of `user_name` for the CIFS share username.

PR #33, opened 2016-02-13
ASM-4774 Add special case for Intel X520/I350 combo card

gavin-scott:
Recognize the Intel X520/I350 combo card as a 2x10Gb,2x1Gb card even
in the case where LinkSpeed is not provided by the iDrac.

Now that ASM passes the chosen NIC FQDD as part of the
network_configuration data this is not strictly necessary, but it's
nice to keep this code in sync with the supported NICs.


gavin-scott:
@asaenz please review.

asaenz:
:+1:

PR #34, opened 2016-02-18
ASM-5633 Refactored nic teaming code for new flexible networking

asaenz:


lib/asm/network_configuration.rb
```diff
@@ -325,13 +337,13 @@ def reset_virt_mac_addr(endpoint)
...
+        raise("NIC MAC Address information needs to updated to network configuration. Invoke nc.add_nics!") unless @network_config_add_nic
+        networks = []
+        mac_teams = {}
         partitions = get_all_partitions
         unless partitions.empty?
           partitions.each do |partition|
-            networks = begin
+            network_objects = begin
               (partition.networkObjects || []).collect do |network|
                 network.staticNetworkConfiguration.delete("ipRange") unless network.staticNetworkConfiguration.nil?
```
gavin-scott: I think this is no longer needed since we do it here:

https://github.com/asaenz/dell-asm-util/blob/abcf345739f22d50f55c94a86036be5db9e9765d/lib/asm/network_configuration.rb#L215

lib/asm/network_configuration.rb
```diff
@@ -113,24 +113,36 @@ def get_all_partitions # rubocop:disable Style/AccessorMethodName
...
     # Finds all networks of one of the specified network types
     def get_networks(*network_types)
+      collect_from_partitions do |partition|
+        (partition.networkObjects || []).find_all do |network|
+          network_types.include?(network.type)
+        end
+      end.flatten.uniq
+    end
+
+    def collect_from_partitions
```
gavin-scott: I like it.

gavin-scott:
Generally looks good other than failing test and I think you can remove that ipRange delete. Would you expect this to work with the existing consumers of the `teams` method? We can merge this first if so if you get the test going.

asaenz:
@gavin-scott I THINK this should work with current teaming code, but to be honest...I'm not 100% sure.

asaenz:
@gavin-scott Addressed your comments, tests are now passing.

asaenz:
Well, tests were passing locally...

PR #37, opened 2016-03-15
ASM-6175 HyperV cluster configuration failed due to disable_pxe getting timed-out

sushilrai:
Increasing LC status check timeout from 5 to 10 minutes to ensure we are waiting enough for server's LC status to be ready after reboot / power-on.

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1153,7 +1153,7 @@ module ASM
     # @option options [FixNum] :timeout (5 minutes) default timeout
     # @return [Hash]
     def poll_for_lc_ready(options={})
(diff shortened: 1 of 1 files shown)


gavin-scott:
Should we also add a power_on at the start of `hyperv_post_installation` to be sure the server is powered on there as well? I'm not sure how failure of the disable_pxe job could result in the server being powered off though.

sushilrai:
We can add power-on to make sure retry works in this scenario. But need to ensure server is fully up and running if we are performing the power-on operation.

gavin-scott:
I think `power_on` is a noop if the server is already on, so I would think that would be safe to run.

PR #38, opened 2016-03-20
ASM-6223 Default to 30 minutes for LC ready timeout

gavin-scott:
This is the default timeout period recommended by the LC team and it
seems like we are still seeing some cases where we time out waiting
for LC ready while disabling PXE boot.


sushilrai:
+1 Do we need to wait for CCB for merging this.

gavin-scott:
I think this should be ok to merge, shouldn't have any negative effects other than causing deployments to take longer to time out.

PR #40, opened 2016-03-28
Asm 6172 2

logankimmel:


gavin-scott:
This is starting to seem pretty hacky to me. I think it would be clearer to just add the "timeout 3600" args in the caller, i.e here:

https://github.com/dell-asm/asm-deployer/blob/e5f762d3b90b2ef908399f07d4145a18660f0f3e/lib/asm/service_deployment.rb#L1062

logankimmel:
fine. revert OG PR and I'll make change to asm-deployer

gavin-scott:
Filed revert PR: https://github.com/dell-asm/asm-deployer/pull/1505

Will merge along with your update when available.

logankimmel:
@gavin-scott I meant the PR for this repo. but I guess because `timeout=nil` it isn't truly necessary

gavin-scott:
Can you add a master / 8.3 PR to remove the timeout argument from `run_command_streaming`?

PR #43, opened 2016-05-06
ASM-6382 Added method to get appliance IP to for default gateway

asaenz:


asaenz:
Will fix/add tests later, but this is good to review otherwise.

PR #44, opened 2016-05-09
ASM-6387 ASM to support Urgent Fury vSAN RA

sushilrai:
As per new recommended flow and Virtual Media needs to be "Attached" state instead of default auto-attach state so that it will always show-up in the device list.
- Execute ConnectRFSISOImage to connect ISO
- Execute PowerStateChange to reboot the host
- Execute GetRemoteServicesAPIStatus, verify LCStatus is Ready
- Execute ChangeBootOrderByInstanceID to set virtual optical as first boot device
- Execute CreateTargetedConfigJob to set the boot order
- Verify config job completes
- Enum DCIM_BootSourceSetting to verify virtual optical is set as first boot device
- Verify system boots to RFS ISO
- Execute DisconnectRFSISOImage to disconnect RFS ISO
- Execute GetRemoteServicesAPIStatus, verify LCStatus is Ready
- Execute ImportSystemConfiguration
- Verify import job completes
- Execute GetRemoteServicesAPIStatus, verify LCStatus is Ready
- Execute ConnectRFSISOImage to connect ISO
- Execute PowerStateChange to reboot the host
- Execute GetRemoteServicesAPIStatus, verify LCStatus is Ready
- Execute ChangeBootOrderByInstanceID to set virtual optical as first boot device
- Execute CreateTargetedConfigJob to set the boot order
- Verify config job completes

### DIFF
--- a/dell-asm-util.gemspec
+++ b/dell-asm-util.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "mocha"
   s.add_development_dependency "puppet"
   s.add_development_dependency "puppetlabs_spec_helper", "0.4.1"
+  s.add_development_dependency "listen", "3.0.7"
 
   s.executables << "wsman_shell.rb"
 
--- a/dell-asm-util.gemspec
+++ b/dell-asm-util.gemspec
@@ -14,6 +14,7 @@ Gem::Specification.new do |s|
   s.add_dependency "nokogiri", "~> 1.5.10"
   s.add_dependency "i18n", "~> 0.6.5"
   s.add_dependency "pry", "~> 0.10"
+  s.add_development_dependency "listen", "3.0.7"
 
   s.add_development_dependency "logger-colors", "~> 1.0.0"
   s.add_development_dependency "guard-shell"
@@ -24,7 +25,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "mocha"
   s.add_development_dependency "puppet"
   s.add_development_dependency "puppetlabs_spec_helper", "0.4.1"
-  s.add_development_dependency "listen", "3.0.7"
 
   s.executables << "wsman_shell.rb"
 
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -19,7 +19,7 @@ module ASM
     POWER_SERVICE = "http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_ComputerSystem?CreationClassName=DCIM_ComputerSystem,Name=srv:system".freeze
     POWER_STATE_CHANGE = "http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_CSPowerManagementService?SystemCreationClassName=DCIM_SPComputerSystem,SystemName=systemmc,CreationClassName=DCIM_CSPowerManagementService,Name=pwrmgtsvc:1".freeze
     SOFTWARE_INSTALLATION_SERVICE = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_SoftwareInstallationService?CreationClassName=DCIM_SoftwareInstallationService,SystemCreationClassName=DCIM_ComputerSystem,SystemName=IDRAC:ID,Name=SoftwareUpdate".freeze
-    IDRAC_CARD_ENUMERATION  = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/DCIM_iDRACCardEnumeration?InstanceID=iDRAC.Embedded.1#VirtualConsole.1#AttachState".freeze
+    IDRAC_CARD_ENUMERATION = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/DCIM_iDRACCardEnumeration?InstanceID=iDRAC.Embedded.1#VirtualConsole.1#AttachState".freeze
     APPLY_ATTRIBUTES = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_iDRACCardService?SystemCreationClassName=DCIM_ComputerSystem,CreationClassName=DCIM_iDRACCardService,SystemName=DCIM:ComputerSystem,Name=DCIM:iDRACCardService".freeze
 
     # rubocop:enable Metrics/LineLength
@@ -1402,7 +1402,6 @@ module ASM
       else
         request_power_state_change(:PowerState => :reboot)
       end
-
 
       # Wait for virtual cd to show up in boot source settings
       max_sleep = 60
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1398,9 +1398,9 @@ module ASM
 
       # Have to reboot in order for virtual cd to show up in boot source settings
       if power_state == :off
-        request_power_state_change(:PowerState => :on)
+        request_power_state_change(:power_state => :on)
       else
-        request_power_state_change(:PowerState => :reboot)
+        request_power_state_change(:power_state => :reboot)
       end
 
       # Wait for virtual cd to show up in boot source settings
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -17,7 +17,11 @@ module ASM
     LC_SERVICE = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_LCService?SystemCreationClassName=DCIM_ComputerSystem,CreationClassName=DCIM_LCService,SystemName=DCIM:ComputerSystem,Name=DCIM:LCService".freeze
     LC_RECORD_LOG_SERVICE = "http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_LCRecordLog?__cimnamespace=root/dcim".freeze
     POWER_SERVICE = "http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_ComputerSystem?CreationClassName=DCIM_ComputerSystem,Name=srv:system".freeze
+    POWER_STATE_CHANGE = "http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_CSPowerManagementService?SystemCreationClassName=DCIM_SPComputerSystem,SystemName=systemmc,CreationClassName=DCIM_CSPowerManagementService,Name=pwrmgtsvc:1".freeze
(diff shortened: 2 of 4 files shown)


ripienaar:
please add to `dell-asm-util.gemspec`

```
s.add_development_dependency "listen", "3.0.7"
```

ripienaar:
@sushilrai ok, now the tests are running but we have lots of failures to fix

asaenz:
@sushilrai Just a minor side note: I actually added the 'listen' dependency in my gemspec to get the tests working for my PR (https://github.com/dell-asm/dell-asm-util/pull/43).

I put it in a bit of a weird spot (my PR got merged before I remembered to fix it), so you might just want to make sure we don't have the same dependency twice in the gemspec file.

lib/asm/wsman.rb
```diff
@@ -751,6 +759,13 @@ def set_attributes(params={}) # rubocop:disable Style/AccessorMethodName
                     :return_value => "0")
     end
 
+    def apply_idrac_attributes(params={})
+      client.invoke("ApplyAttributes ", APPLY_ATTRIBUTES,
```
gavin-scott: Looks like there is a trailing space here before the closing `"`

lib/asm/wsman.rb
```diff
@@ -982,6 +997,19 @@ def set_power_state(params={}) # rubocop:disable Style/AccessorMethodName
                     :return_value => "0")
     end
 
+    # Update power state of the server.
+    #
+    # @param options [Hash]
+    # @option options [Symbol|String] :power_state 1 to 10"
+    # @return [Hash]
+    # @raise [ResponseError] if the command fails
+    def request_power_state_change(params={})
```
gavin-scott: What's the difference between this and `set_power_state`? -- https://github.com/sushilrai/dell-asm-util/blob/0d6d3b75564a8e712505176ed9ed5c67da6704b3/lib/asm/wsman.rb#L993

lib/asm/wsman.rb
```diff
@@ -1269,13 +1297,25 @@ def set_boot_order(boot_device, options={})
 
       change_boot_order_by_instance_id(:instance_id => "IPL",
                                        :source => target[:instance_id])
-      change_boot_source_state(:instance_id => "IPL", :enabled_state => "1",
-                               :source => target[:instance_id])
```
gavin-scott: You mentioned this to me in the call earlier. But why don't we need to make sure that the virtual CD is enabled?

lib/asm/wsman.rb
```diff
@@ -225,6 +229,10 @@ def boot_source_settings
       client.enumerate("http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_BootSourceSetting?__cimnamespace=root/dcim")
     end
 
+    def virtual_media_attach_status
```
gavin-scott: Maybe make this a more generic iDRACCardEnumeration method since when you call it later you search through the results for the AttachState attribute anyway?

lib/asm/wsman.rb
```diff
@@ -1269,13 +1297,25 @@ def set_boot_order(boot_device, options={})
...
-      change_boot_source_state(:instance_id => "IPL", :enabled_state => "1",
-                               :source => target[:instance_id])
       run_bios_config_job(:target => boot_mode[:fqdd],
                           :scheduled_start_time => options[:scheduled_start_time],
                           :reboot_job_type => options[:reboot_job_type])
     end
 
+    def set_virtual_media_attach_state
+      va = virtual_media_attach_status.find { |x| x[:attribute_display_name] == "Attach State" }
+      if va[:current_value] == "Auto-Attach"
```
gavin-scott: What about if it's Detached?

lib/asm/wsman.rb
```diff
@@ -982,6 +997,19 @@ def set_power_state(params={}) # rubocop:disable Style/AccessorMethodName
...
+    # Update power state of the server.
+    #
+    # @param options [Hash]
+    # @option options [Symbol|String] :power_state 1 to 10"
+    # @return [Hash]
+    # @raise [ResponseError] if the command fails
+    def request_power_state_change(params={})
+      client.invoke("RequestPowerStateChange", POWER_STATE_CHANGE,
+                    :params => params,
+                    :required_params => :PowerState,
```
gavin-scott: Let's use snake case for the parameters, i.e. `:power_state`

lib/asm/wsman.rb
```diff
@@ -1309,10 +1349,18 @@ def boot_rfs_iso_image(options={})
...
+      set_virtual_media_attach_state
+
       connect_rfs_iso_image(options)
 
       # Have to reboot in order for virtual cd to show up in boot source settings
-      reboot(options)
+      if power_state == :off
+        request_power_state_change(:PowerState => "2")
+      else
+        request_power_state_change(:PowerState => "10")
```
gavin-scott: You can use descriptive symbols instead of the magic values 2 and 10 by adding a `:power_state` case here:

https://github.com/dell-asm/dell-asm-util/blob/364d140a2edd1549ce5d6a6443b9f171c43c6a14/lib/asm/wsman/parser.rb#L144

lib/asm/wsman.rb
gavin-scott: In either case, please add yard docs and a unit test.

lib/asm/wsman.rb
```diff
@@ -751,6 +759,13 @@ def set_attributes(params={}) # rubocop:disable Style/AccessorMethodName
                     :return_value => "0")
     end
 
+    def apply_idrac_attributes(params={})
```
gavin-scott: Please add YARD docs and unit test.

lib/asm/wsman.rb
```diff
@@ -1269,13 +1297,25 @@ def set_boot_order(boot_device, options={})
...
       change_boot_order_by_instance_id(:instance_id => "IPL",
                                        :source => target[:instance_id])
-      change_boot_source_state(:instance_id => "IPL", :enabled_state => "1",
-                               :source => target[:instance_id])
       run_bios_config_job(:target => boot_mode[:fqdd],
                           :scheduled_start_time => options[:scheduled_start_time],
                           :reboot_job_type => options[:reboot_job_type])
     end
 
+    def set_virtual_media_attach_state
```
gavin-scott: Please add YARD docs and unit test.

lib/asm/wsman.rb
sushilrai: WSMAN Service used are different. WSMAN endpoint used for method "request_power_state_change" is new endpoint recommended by LC team

lib/asm/wsman.rb
sushilrai: If we have vmedia set to attach, virtual optical will always be listed in the boot order

gavin-scott: Yeah, but it could still be _disabled_ I think. I.e. if you go into the boot settings screen in idrac console it could be unchecked.

lib/asm/wsman.rb
gavin-scott: OK, can you either remove the old method then if that seems feasible or mark it as deprecated in the YARD docs with a reference to the new method callers should use? http://www.rubydoc.info/gems/yard/file/docs/Tags.md#deprecated

lib/asm/wsman.rb
sushilrai: We are setting the Virtual Media to attached state before this code is executed. That will take care of adding the Virtual Media to boot device list.

lib/asm/wsman.rb
sushilrai: This method is a work-around fix and hopefully power-cycle changes will be fixed in 2.40.40 (as per LC team).

aimers1975:
@sushilrai and @gavin-scott  Same question on this PR, is this needed and going in?  If so, it needs to be in by Sunday...Sushil, please work with Gavin to resolve any remaining issues

gavin-scott:
@sushilrai I fixed the rubocop failures with this commit in my repo: https://github.com/dell-asm/dell-asm-util/commit/e783a07ba71e868782db092fba45a0a7a57fe89b  Can you fetch my repo and cherry-pick that in and add to this PR?

lib/asm/wsman.rb
gavin-scott: Setting it to attach mode just means that the device will always be present in the boot list. It doesn't mean that it will always be enabled. So I think we still should ensure that the virtual CD is enabled.

Here is a screenshot showing the Attach mode Virtual Optical Drive as disabled:

image

lib/asm/wsman.rb
```diff
@@ -1309,10 +1391,18 @@ def boot_rfs_iso_image(options={})
       options = {:reboot_job_type => :graceful_with_forced_shutdown,
                  :reboot_start_time => "TIME_NOW",
                  :timeout => 10 * 60}.merge(options)
+      # Virtual Media needs to be in attach state
+      set_virtual_media_attach_state
+
       connect_rfs_iso_image(options)
 
       # Have to reboot in order for virtual cd to show up in boot source settings
-      reboot(options)
```
gavin-scott: This still worries me. There are other places we use `CreateRebootJob`, e.g. in dell-idrac firmware update or asm-deployer calls the `wsman.reboot` method. Do we need to change them all? Should we just change the `reboot` method to use the code you have below instead of special-casing it here?

dell-asm-util.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "mocha"
   s.add_development_dependency "puppet"
   s.add_development_dependency "puppetlabs_spec_helper", "0.4.1"
+  s.add_development_dependency "listen", "3.0.7"
```
gavin-scott: @asaenz thinks this is going to cause problems when merged because it's already in 8.2.1 on line 17: https://github.com/dell-asm/dell-asm-util/blob/652775e9fbe9330bd5b159f3294808714fe35920/dell-asm-util.gemspec#L17

Can you move it to line 17?

gavin-scott:
After reviewing the jira around idrac CreateRebootJob problems, I'm not convinced that it described a problem that we have ever seen in one of these code flows. I am able to execute the `reboot` method without problems on latest firmwares.

In light of that I'm not convinced the changes here make any real difference in the original problem (*) -- occasional hangs of BIOS CreateTargetedConfigJobs seen on latest idrac firmware. We need to continue to try to root cause the problem with the LC team.

(*) - using attach instead of auto-attach seems like a good optimization for our flow (eliminates one server reboot), but I'd rather not introduce that change for 8.2.1 if possible.

gavin-scott:
Going to go ahead and merge since we need an 8.2.1 build to start testing with and the VSAN AHCI flow does work better with these changes. Still want to root cause the original problem that this seems to be working around, but will do that separately.

PR #46, opened 2016-05-11
ASM-6656 Wsman invoke input param

logankimmel:
Allow `:input_file` param to be passed into ASM::WsMan::Client#invoke


gavin-scott:
You got rubocopped. Other than that :+1:

PR #48, opened 2016-05-14
ASM-6677 Power-off server and razor policy bind/reboot race condition

sushilrai:
Added power status check and retry of the status check to ensure server is powered off before success status is returned.


gavin-scott:
See my e-mail but I don't think this is a problem. The problem in your deployment was that the ImportSystemConfiguration job seems to have failed to set the boot order properly.

We may want to look at adding a `power_on` in the OS installation flow to ensure that the server is on after we set PXE first in the boot order.

PR #50, opened 2016-05-17
ASM-6558-iDrac 2.40.40.40 requires TLS 1.2 by default

Shreya99:
Upgrading wsmancli 2.6.0 to support TLS changes in the next idrac firmware 2.40.40.40

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -11,13 +11,13 @@ require "uri"
 module ASM
   class WsMan
     # rubocop:disable Metrics/LineLength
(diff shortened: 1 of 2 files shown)


ripienaar:
was this broken before or is this just a change to make it correct?

This is the kind of information I'd expect to learn when reading the commit message, please update the PR with a better commit message

gavin-scott:
The change is to work with the new wsmancli 2.6.0. We need to upgrade to that to support TLS changes that will be enforced in the next idrac firmware 2.40.40.40 expected to be released ~ September.

@Shreya99 can you add the jira to the title of this PR and the commit and update the commit message and PR description to include the information I give above?

Also it would be good to add some error handling so that if this code is used with an older version of wsmancli we output a sensible error message letting the user know that there version of wsmancli is not compatible with this library. This would probably be the right place to add that kind of error handling:

https://github.com/dell-asm/dell-asm-util/blob/e573a1c065cd176b8e8283340a501eb9a60b191d/lib/asm/wsman/client.rb#L100

gavin-scott:
Closing in favor of #59

PR #51, opened 2016-05-17
ASM-6387 Boot RFS ISO image workflow tweaks

gavin-scott:
Recent changes were made to the boot_rfs_iso_image workflow in
response to problems that were seen where ImportSystemConfiguration or
CreateTargetedConfig jobs became stuck in running state. LC team has
now root caused this and the problem was arising when an
ImportSystemConfiguration job was run that set BiosBootSeq or HddSeq
but the applied configuration did not result in any actual
configuration changes. In that case subsequent jobs would become stuck
in Running state.

In response to the root cause analysis this commit adds back some
things that were removed previously:

- Revert back to using CreateRebootJob in boot_rfs_iso_image as that
  was not a source of problems.

- Only reboot the server after connecting the RFS ISO image if the
  virtual CD is not already detected. In Attach mode the virtual CD
  device will always be available but in Auto-Attach mode we will
  still need to reboot.

- Add back the ChangeBootSourceState call to ensure that the target
  boot device is enabled.

- Allow use of auto-attach mode if desired. Using attach mode was
  suggested by the LC team and is a good optimization as it will allow
  the workflow to skip a reboot if set, but we shouldn't require it.

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1334,22 +1334,40 @@ module ASM
 
       change_boot_order_by_instance_id(:instance_id => "IPL",
                                        :source => target[:instance_id])
+
+      unless target[:current_enabled_status] == "1"
+        logger.info("Enabling boot device %s on %s" % [target[:instance_id], host])
+        change_boot_source_state(:instance_id => "IPL", :enabled_state => "1",
+                                 :source => target[:instance_id])
+      end
+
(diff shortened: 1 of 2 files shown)


gavin-scott:
DO NOT MERGE.

Ready for review, but doing end-to-end testing.

aimers1975:
@asaenz Here is the PR for you to review sometime today...

asaenz:
👍

PR #53, opened 2016-05-20
ASM-6729 Ensure boot_rfs_iso_image reboots server

gavin-scott:
With the change to use Attach mode instead of Auto-Attach by default
(where the virtual CD is always listed in the boot order even if no
virtual CD is connected), it's possible that when boot_rfs_iso_image
is called the virtual CD is already first in the boot order. In that
case there was no need to change the boot order and the server did not
get rebooted into the newly attached ISO. In that case an explicit
reboot is now issued.


gavin-scott:
DO NOT MERGE -- still end-to-end testing.

PR #60, opened 2016-06-28
ASM-6515 ASM deployment to support EMC Storage provisioning for HyperV FC

sushilrai:
Added new helped method to retrieve WWPN and WWNN for the servers. This is required for EMC host authentications where tuple of WWPN and WWNN is required instead of just WWPN

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -330,6 +330,16 @@ module ASM
       end.compact
     end
 
+    # Retrieve WWNN and WWPN for server using NIC_FC_View
+    #
+    # @param endpoint [Hash] Hash of server access endpoint
+    # @param logger [Object] Logger object
+    #
+    # @return [Array<Array>]
+    #
+    # @example return
+    #   [["20:00:00:24:FF:4A:BB:5A", "21:00:00:24:FF:4A:BB:5A"],
+    #   ["20:00:00:24:FF:4A:BB:5B", "21:00:00:24:FF:4A:BB:5B"]]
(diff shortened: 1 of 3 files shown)


lib/asm/wsman.rb
```diff
@@ -330,6 +330,13 @@ def self.get_wwpns(endpoint, logger=nil)
       end.compact
     end
 
+    def self.get_wwpns_wwnns(endpoint, logger=nil)
+      response = invoke(endpoint, "enumerate",
+                        "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/DCIM/DCIM_FCView",
+                        :logger => logger)
+      ( response.scan(/<n1:VirtualWWN>(\S+)<\/n1:VirtualWWN>\s*<n1:VirtualWWPN>(\S+)<\/n1:VirtualWWPN>/) || [] )
+    end
```
ripienaar: tests and docs please, plenty of examples in this file and its spec

spec/unit/asm/wsman_spec.rb
```diff
@@ -764,4 +762,17 @@
       wsman.boot_to_network_iso_image
     end
   end
+
+  describe "#get_wwpn_wwnn" do
+    before do
+      @wwpn_response = SpecHelper.load_fixture("wsman/fc_view.xml")
+    end
```
ripienaar: replace this before block with `let(:wwpn_response) { SpecHelper..... }`, we've been using lets pretty much all over for over a year I think and its how the rest of this file looks too

PR #61, opened 2016-07-06
ASM-7035 VSAN deployment fails with VSAN partition on back side disk

sushilrai:
For some servers, power-off operation is taking some time. Added validation loop to ensure server is powered-off when control is moved out of the method

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1073,6 +1073,10 @@ module ASM
 
       if power_state != :off
         set_power_state(:requested_state => :off)
+        (1..30).each do |wait_counter|
+          break if power_state == :off
+          sleep 10
+        end
       else
         logger.debug "Server is already powered off"
       end
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1065,6 +1065,8 @@ module ASM
 
     # Power the server off.
     #
+    # After executing the power-off operation, check the power-state and returns once server is in power-off state
+    #
     # @return [void]
     # @raise [ResponseError] if the command fails
     def power_off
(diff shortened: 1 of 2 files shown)


lib/asm/wsman.rb
```diff
@@ -1073,6 +1073,10 @@ def power_off
 
       if power_state != :off
         set_power_state(:requested_state => :off)
+        (1..30).each do |wait_counter|
+          break if power_state == :off
+          sleep 10
+        end
```
ripienaar: we should log if we get here and its still not off? And probably log while we're trying too?

aimers1975:
@sushilrai is it expected that the tests are failing.. @asaenz are you able to review despite the failing tests.  Note this is an urgent 8.2.1 fix.

lib/asm/wsman.rb
```diff
@@ -1073,6 +1075,11 @@ def power_off
 
       if power_state != :off
         set_power_state(:requested_state => :off)
+        (1..30).each do |wait_counter|
+          break if power_state == :off
+          logger.debug "Server is not in power-off state. Retry counter %d" % wait_counter if logger
+          sleep 10
+        end
```
ripienaar: I am still wondering what happens at the end of the retry loop if the machine is still on? 

To me this should raise an error - and the docs set that expectation by stating a exception is raised when the command fails.

Would that make sense? As it is now the machine can still be on and no-one would be told, hard to say whats best since that will be a behaviour change so we could possibly do that in another ticket, @gavin-scott any ideas?

sushilrai: Don't want to change the behaviour in current code state thats why skipped raising the error

ripienaar: OK, agree lets see what gavin says and we can make a new ticket.  BUT can you add a logger.warn telling that it failed for now? no behaviour change but its not completely invisible then?

PR #62, opened 2016-07-08
ASM-7091 Use correct attribute for attach mode

gavin-scott:
The original code for setting attach mode was setting the incorrect
iDrac card enumeration enumeration attribute. This meant that the
server was always left in auto attach mode which requires an
additional reboot in order for the virtual CD to show up in the boot
order list after we attach the ISO. In attach mode the virtual CD is
always present in the boot order list even if no ISO is attached.

In addition to slowing down the install process the additional reboot
was occasionally causing servers to get rebooted while an OS
installation was already in progress and corrupting that install.

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1367,16 +1367,16 @@ module ASM
     # @param state [Symbol] :detach, :attach or :auto_attach
     # @return [void]
     def set_virtual_media_attach_state(state) # rubocop:disable Style/AccessorMethodName
-      state_map = {:detached => "Detached", :attached => "Attached", :auto_attach => "Auto-Attach"}
+      state_map = {:detached => "Detached", :attached => "Attached", :auto_attach => "AutoAttach"}
       state_string = Parser.enum_value(nil, state_map, state)
 
-      va = idrac_card_enumeration.find { |x| x[:attribute_display_name] == "Attach State" }
+      va = idrac_card_enumeration.find { |x| x[:attribute_display_name] == "RAC Virtual Media Attached" }
       if !va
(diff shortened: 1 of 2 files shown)


lib/asm/wsman.rb
```diff
@@ -1367,16 +1367,16 @@ def set_boot_order(boot_device, options={})
     # @param state [Symbol] :detach, :attach or :auto_attach
     # @return [void]
     def set_virtual_media_attach_state(state) # rubocop:disable Style/AccessorMethodName
-      state_map = {:detached => "Detached", :attached => "Attached", :auto_attach => "Auto-Attach"}
+      state_map = {:detached => "Detached", :attached => "Attached", :auto_attach => "AutoAttach"}
```
gavin-scott: Note this attribute uses the string `AutoAttach` instead of `Auto-Attach`.

lib/asm/wsman.rb
```diff
@@ -1367,16 +1367,16 @@ def set_boot_order(boot_device, options={})
...
       state_string = Parser.enum_value(nil, state_map, state)
 
-      va = idrac_card_enumeration.find { |x| x[:attribute_display_name] == "Attach State" }
+      va = idrac_card_enumeration.find { |x| x[:attribute_display_name] == "RAC Virtual Media Attached" }
```
gavin-scott: This PR is for the `release/dell-asm-8.2.0` branch only. I'd like to rework for the master branch as it seems a little screwy that we match based on `attribute_display_name` here but then we have to set it with the `attribute_name` below on line 1379. Seems like we should match on the `attribute_name` as well to avoid duplication.

PR #63, opened 2016-07-11
 misc: fix dependant builds

ripienaar:
Pins restclient to a version that works with our outdated ruby


ripienaar:
@sushilrai I think this will fix the failures in https://github.com/dell-asm/asm/pull/313

PR #65, opened 2016-07-18
ASM-7203 Added `ASM::Util#deep_merge!`

logankimmel:


ripienaar:
rubocop fails on the tests, can probably fix it with `rubocop -a foo.rb`

PR #66, opened 2016-08-08
ASM-7202:added method to get if there is fc interface_type

arunrordell:


arunrordell:
@gavin-scott  test are failing because of the gem shall i submit a PR to include Gem?

ripienaar:
in this PR - add `s.add_development_dependency "json_pure", "2.0.1"` to the gemspec file

lib/asm/network_configuration.rb
```diff
@@ -166,14 +168,15 @@ def get_static_ips(*network_types)
 
     def build_cards(interfaces)
       return [] unless interfaces
-
```
gavin-scott: Minor but can you add this space back in, I think it's easier to read with it.

PR #67, opened 2016-08-11
ASM-7468 Rescue non-UTF-8 wsman response

logankimmel:
Reset the iDRAC when a wsman call returns non-UTF-8 encoding

### DIFF
--- a/lib/asm/racadm.rb
+++ b/lib/asm/racadm.rb
@@ -1,0 +1,34 @@
+require "asm/racadm/transport"
+
+module ASM
+  class IdracResetError < Exception; end
+  class Racadm
+    attr_reader :transport, :endpoint, :logger
+
+    def initialize(endpoint, options={})
+      @logger = ASM::Util.augment_logger(options.delete(:logger) || Logger.new(nil))
+      @transport = ASM::Racadm::Transport.new(endpoint, options)
+      @endpoint = endpoint
+    end
+
+    def client
+      transport.client
+    end
+
+    def reset_idrac
+      logger.info("Resetting IDrac...")
+
+      client.exec "racadm racreset soft" do |_, stream, data|
+        logger.debug(data)
+
+        # Issue warning for the message 'Could not chdir to home directory /flash/data0/home/root: No such file or directory' else raise error
+        if data.include? "Could not chdir to home directory"
+          logger.warning "Warning for message - #{data}"
+        elsif stream == :stderr
+          raise IdracResetError
+        end
+      end
+      client.close
+    end
+  end
+end
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -5,6 +5,7 @@ require "asm/util"
 require "asm/wsman/client"
 require "asm/wsman/parser"
 require "asm/wsman/response_error"
+require "asm/racadm"
 require "rexml/document"
 require "uri"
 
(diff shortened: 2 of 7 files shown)


lib/asm/racadm.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        #Issue warning for the message 'Could not chdir to home directory /flash/data0/home/root: No such file or directory' else raise error
+        if data.include? "Could not chdir to home directory"
+          logger.warning "Warning for message - #{data}"
+        elsif stream == :stderr
+          raise IdracResetError
+        end
+      end
+      client.close
+      ASM::WsMan.new(endpoint, :logger => logger).poll_for_lc_ready(:timeout => 360)
```
gavin-scott: I would move this out into the calling code to void unnecessary cross-dependencies between ASM::WsMan and ASM::Racadm. Also `poll_for_lc_ready` doesn't handle the case of the iDrac being down due to reset, it will error out pretty quickly. It's really intended just to poll the `GetRemoteServicesAPIStatus`

lib/asm/racadm.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+  end
+end
```
gavin-scott: EOL

lib/asm/wsman.rb
```diff
@@ -1260,7 +1271,7 @@ def poll_for_lc_ready(options={})
       detach_iso_image if resp["iso_attach_status"] == "1"
 
       max_sleep_secs = 60
-      resp = ASM::Util.block_and_retry_until_ready(options[:timeout], RetryException, max_sleep_secs) do
+      resp = ASM::Util.block_ and_retry_until_ready(options[:timeout], RetryException, max_sleep_secs) do
```
gavin-scott: You got a space in `block_and_retry_until_ready`

lib/asm/wsman.rb
```diff
@@ -226,7 +227,17 @@ def boot_config_settings
     #       :pending_enabled_status=>"1"}, # ...
     #     ]
     def boot_source_settings
+      tries ||= 1
       client.enumerate("http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_BootSourceSetting?__cimnamespace=root/dcim")
+    rescue ASM::WsMan::ResponseError
+      if tries == 1
+        logger.debug("Parser error retrieving DCIM_BootSourceSetting due to non regular UTF-8. Resetting iDRAC and retrying.")
```
gavin-scott: I'd rather handle the specific UTF-8 ResponseError rather than all ResponseErrors here with retry.

Also it might be better to hoist the reset and retry out into the asm-deployer layer -- it's already doing retry of the `boot_rfs_iso_image` call

gavin-scott:
Tests are failing -- can't load `net/ssh`

logankimmel:
@gavin-scott Updated everything without moving code into asm-deployer.  If this needs to be done I'll have to get it tomorrow

gavin-scott:
- Tests are still failing
- Still concerned about `poll_for_lc_ready` not handling iDrac being offline temporarily
- Moving the retry to asm-deployer not strictly necessary, seems a bit better to me though

This doesn't have to be merged tonight, we can wait until tomorrow.

logankimmel:
@ripienaar any ideas on on this net-ssh failure?

gavin-scott:
Don't you just need to add it to the .gemspec?

gavin-scott:
It says `Gem::InstallError: net-ssh requires Ruby version >= 2.0.`. See if you can pin it to a compatible version.

logankimmel:
Good find, I think this is ready now.  I don't really like the idea of updating asm-deployer to handle the retry because we've only seen this happen on that particular call. If the UTF-8 issue creeps into other wsman calls than, yeah, I think it's reasonable.

gavin-scott:
Fair enough, looks pretty good to me.

PR #68, opened 2016-08-13
[maint] Rename has_fc to has_fc?

gavin-scott:
Generally boolean accessors should end with a question mark


ripienaar:
we're sure nothing uses this? 👍  otherwise

gavin-scott:
Yeah, nothing uses it currently, it was added very recently as prep work for ASM-7202

PR #73, opened 2016-08-29
ASM-7739 Remove Mssql passwords as well from log

gavin-scott:
Some of msslq password keys are in the short form
of password ex: SApwd, addition pwd to the condition
such that pwd fields also replaced with *


gavin-scott:
8.2.1 backport of #72

PR #74, opened 2016-08-29
ASM-7739 Remove Mssql passwords as well from log

shaorong-zhu:
Some of msslq password keys are in the short form
of password ex: SApwd, addition pwd to the condition
such that pwd fields also replaced with *

(cherry picked from commit 2ff965f42ff498a8c5f34a65c1a1c6bbd224bbec)

@gavin-scott Please review and merge. Thanks


gavin-scott:
Sorry, already got this in as #73

PR #75, opened 2016-10-06
ASM-7884 migrating firmware update utility from Idrac module to util

prem548:
migrating firmware update IDRAC module to Util. through which we  can remove possible unnecessary XML config files in IDRAC module , By using only existing wsman commands. 

cautions: 
     Had to include "net/ssh" library to Gem file.

### DIFF
--- a/lib/asm/firmware.rb
+++ b/lib/asm/firmware.rb
@@ -1,0 +1,333 @@
+require "asm/wsman"
+require "asm/wsman/client"
+require "erb"
+require "tempfile"
+require "asm/transport/racadm"
+
+module ASM
+  class Firmware
+    IDRAC_ID = 252_27
+    LC_ID = 288_97
+    UEFI_DIAGNOSTICS_ID = 258_06
+    DRIVER_PACK = 189_81
+    OS_COLLECTOR = 101_734
+
+    # Component ids that do not require a reboot (DIRECT UPDATES)
+    NO_REBOOT_COMPONENT_IDS = [IDRAC_ID, LC_ID, UEFI_DIAGNOSTICS_ID, DRIVER_PACK, OS_COLLECTOR].freeze
+
+    # Max time to wait for a job to complete
+    MAX_WAIT_SECONDS = 1800
+    attr_reader :logger
+
+    def initialize(cred, options={})
+      @endpoint = cred
+      @logger = options[:logger]
+    end
+
+    # Used to update server firmware via iDrac
+    #
+    # @param config [Hash] schema of resources used to update the firmware
+    # @example
+    #        {"asm::server_update"=>
+    #          {"rackserver-5xclw12"=>
+    #            {"asm_hostname" =>"172.25.5.100",
+    #             "force_restart" => false,
+    #             "install_type" => "uri",
+    #             "path" => "/var/nfs/firmware/ff8080815746abab0157481ecea1000f/ASMCatalog.xml",
+    #             "server_firmware" => "[{\"instance_id\":\"DCIM:INSTALLED#701__NIC.Embedded.1-1-1\",
+    #                                     \"uri_path\":\"/var/nfs/firmware/ff8080815746abab0157481ecea1000f/FOLDER03287319M/3/Network_Firmware_0MT4K_WN64_7.10.64.EXE\"
+    #                                   }]"
+    #            }
+    #          }
+    #        }
+    # @param resource_hash [Array<Hash>] of instances with nfs and uri paths
+    # @example
+    #      [{"instance_id" => "DCIM:INSTALLED#701__NIC.Embedded.1-1-1",
+    #        "uri_path" => "nfs://172.25.5.100/FOLDER03287319M/3/Network_Firmware_0MT4K_WN64_7.10.64.EXE;mountpoint=/var/nfs/firmware/ff808081578bd88601578d525d4e004e"
+    #       },
+    #       {"instance_id" => "DCIM:INSTALLED#iDRAC.Embedded.1-1#IDRACinfo",
+    #        "component_id" => "25227",
+    #        "uri_path" => "nfs://172.25.5.100/FOLDER03526343M/2/iDRAC-with-Lifecycle-Controller_Firmware_JHF76_WN64_2.30.30.30_A00.EXE;
+    #                       mountpoint="/var/nfs/firmware/ff808081578bd88601578d525d4e004e"
+    #       }]
+    # @param device_config [Hash]
+    # @example
+    #      {"cert_name"=>"rackserver-5xclw12",
+    #       "host"=>"172.17.5.35",
+    #       "port"=>nil,
+    #       "path"=>"/asm/bin/idrac-discovery.rb",
+    #       "scheme"=>"script",
+    #       "arguments"=>{"credential_id"=>"ff80808157bfd05a0157bfd13392000d"},
+    #       "user"=>"username",
+    #       "enc_password"=>nil,
+    #       "password"=> "1234"
+    #      }
+    #
+    # @param logger [Logger]
+    # @return [void]
+    # @raise [Error] resource_hash was empty
+    def self.idrac_fw_install_from_uri(config, resource_hash, device_config, logger)
+      raise("Received empty resources to update the firmware on server") if resource_hash.nil?
+
+      cert_name = config["asm::server_update"].keys.first
+      options = {:host => device_config["host"], :user => device_config["user"], :password => device_config["password"]}
+      firmware_instance = ASM::Firmware.new(options, :logger => logger)
+      wsman_instance = ASM::WsMan.new(options, :logger => logger)
+      firmware_instance.clear_job_queue_retry(wsman_instance)
+      force_restart = config["asm::server_update"][cert_name]["force_restart"]
+      pre = []
+      main = []
+
+      resource_hash.each do |firmware|
+        logger.debug(firmware)
+        if [LC_ID, IDRAC_ID].include? firmware["component_id"].to_i
+          pre << firmware
+        else
+          main << firmware
+        end
+      end
+
+      unless pre.empty?
+        logger.debug("LC Update required, installing first")
+        firmware_instance.update_idrac_firmware(pre, force_restart, wsman_instance)
+      end
+      firmware_instance.update_idrac_firmware(main, force_restart, wsman_instance)
+
+      # After updating Ensure LC is up and in good state before exiting
+      wsman_instance.poll_for_lc_ready
+    end
+
+    # Clearing the job_queue with retry
+    #
+    # make sure lc ready after clearing the job queue other wise reset the IDRAC
+    # @param wsman [Object]
+    # @return [void]
+    # @raise [StandardError] if the command delete_job_queue operation or lc_status was not ready after job queue deletion
+    def clear_job_queue_retry(wsman)
+      attempts ||= 0
+      begin
+        attempts += 1
+        logger.debug("Clearing the Job Queue...")
+        resp = wsman.delete_job_queue(:job_id => "JID_CLEARALL")
+
+        if resp[:return_value] == "0"
+          logger.debug("Jobs in job queue was cleared successfully...")
+        else
+          raise("Unable to clear all the job queue")
+        end
+
+        wsman.poll_for_lc_ready # Make sure Lc status was ready
+      rescue StandardError => e
+        logger.debug("Job queue cannot be cleared.") if attempts > 1
+        logger.debug("Job queue still shows jobs exist after attempting to clear the job queue.")
+        logger.debug("Caught exception in clearing job queue %s : %s" % [e, e.class])
+        logger.debug("Resetting the Idrac ...")
+
+        # Resets two times
+        if attempts < 3
+          transport = ASM::Transport::Racadm.new(ASM::WsMan::Client.endpoint, logger)
+          transport.reset_idrac
+          sleep 60
+          retry
+        else
+          raise("Unable to find the LC status after clearing job queue")
+        end
+      end
+    end
+
+    # Used to update the iDrac firmware
+    #
+    # @param firmware_list [Array[<Hash>]] Each instance with mount and nfs path
+    # @param force_restart [Boolean]
+    # @param wsman [Object]
+    # @return [void]
+    # @raise [StandardError] if firmware gets_install_uri_job was not able to create job_id to receive or created duplicate job_ids or firmware update fails
+    def update_idrac_firmware(firmware_list, force_restart, wsman)
+      statuses = {}
+
+      # Initiate all firmware update jobs
+      firmware_list.each do |fw|
+        logger.debug(fw)
+        job_id = gets_install_uri_job(fw, wsman)
+        raise("Failed to initiate the firmware job for %s" % fw) unless job_id
+        raise("Duplicate job id %s for firmware %s: %s" % [job_id, fw, statuses[job_id]]) if statuses[job_id]
+        statuses = block_until_downloaded(job_id, fw, wsman)
+      end
+
+      statuses.each do |_, v|
+        if NO_REBOOT_COMPONENT_IDS.include?(v[:firmware]["component_id"].to_i)
+          v[:desired] = "Completed"
+          v[:reboot_required] = false
+        else
+          force_restart ? v[:desired] = "Completed" : v[:desired] = "Scheduled"
+          v[:reboot_required] = true
+        end
+      end
+
+      reboot_firmwares = statuses.select { |_, v| v[:reboot_required] }
+      completed_endstate_firmwares = statuses.select { |_, v| v[:desired] == "Completed" }
+      scheduled_endstate_firmwares = statuses.select { |_, v| v[:desired] == "Scheduled" }
+
+      schedule_reboot_job_queue(reboot_firmwares, force_restart, wsman) # Reboot required if firmware instance does not include :component_id key
+
+      [scheduled_endstate_firmwares, completed_endstate_firmwares].each do |firmware_set|
+        until firmware_set.values.all? { |v| v[:status] =~ /#{v[:desired]}|Failed|InternalTimeout/ }
+          firmware_set.each do |key, val|
+            if val[:status] != val[:desired] && Time.now - val[:start_time] > MAX_WAIT_SECONDS
+              val[:status] = "InternalTimeout"
+            else
+              lc_job = wsman.get_lc_job(key)
+              val[:status] = lc_job[:job_status]
+              logger.debug("Job Status #{key}: #{val[:status]}")
+            end
+          end
+          sleep 30
+        end
+      end
+      # Raise an error if any firmware jobs failed
+      failures = statuses.values.find_all { |v| v[:status] =~ /Failed|InternalTimeout/ }
+      if failures.empty?
+        logger.debug("Firmware update completed successfully")
+      else
+        logger.info("Failed firmware jobs: #{failures}")
+        raise("Firmware update failed in the lifecycle controller. Please refer to LifeCycle job logs")
+      end
+    end
+
+    # Used to install specified firmware instance on IDRAC
+    #
+    # @param firmware [Hash]
+    # @param wsman [Object]
+    # @option firmware [String] :instance_id specifies the particular firmware instance ID
+    # @option firmware [String] :uri_path specifies NFS mount share path
+    # @option firmware [String] :component_id specifies the component_id for iDRAC
+    # @return [String] created JOB_ID
+    # @raise [ASM::Error] if install_from_uri wsman command fails or unable to mount the specified firmware path
+    def gets_install_uri_job(firmware, wsman)
+      config_file = create_xml_config_file(firmware["instance_id"], firmware["uri_path"])
+      resp = wsman.install_from_uri(:input_file => config_file.path)
+      if resp[:return_value] == "4096"
+        job_id = resp[:job]
+        logger.debug("InstallFromURI started")
+        logger.debug("JOB_ID: #{job_id}")
+      else
+        logger.debug("Error installing From URI config: #{config_file.read}")
+        raise("Problem running InstallFromURI: #{resp[:message]}")
+      end
+      job_id
+    end
+
+    # Helper method used to block till firmware status completed
+    #
+    # @param job_id [String]
+    # @param firmware [Hash]
+    # @param wsman [Object]
+    # @option firmware [String] :instance_id defines each firmware instance ID
+    # @option firmware [String] :uri_path defines specific mount share path
+    # @option firmware [String] :component_id by default component id is "25227"
+    # @return statuses [Hash]
+    # @option statuses [String] :job_id created job_id
+    # @option statuses ["new", "Failed", "TemporaryFailure", "Downloaded", "Completed"] :status statuses
+    # @option statuses [String] :firmware defines specific firmware instance
+    # @raise [StandardError]  if the checkjobstatus command fails
+    # @raise [Error] if the firmware job status was failed
+    def block_until_downloaded(job_id, firmware, wsman)
+      statuses = {
+        job_id => {
+          :job_id => job_id,
+          :status => "new",
+          :firmware => firmware,
+          :start_time => Time.now
+        }
+      }
+      until statuses[job_id][:status] =~ /Downloaded|Completed|Failed/
+        sleep 30
+        begin
+          lc_status = wsman.get_lc_job(job_id)
+          statuses[job_id][:status] = lc_status[:job_status]
+        rescue StandardError => e
+          statuses[job_id][:status] = "TemporaryFailure"
+          logger.warn("Look up job status #{job_id} failed: #{e}")
+        end
+        logger.debug("Job Status: #{statuses[job_id][:status]}")
+
+        if Time.now - statuses[job_id][:start_time] > MAX_WAIT_SECONDS
+          logger.warn("Timed out waiting for firmware job #{job_id} to complete")
+          statuses[job_id][:status] = "Failed"
+        end
+      end
+
+      if statuses[job_id][:status] == "Completed"
+        logger.debug("Firmware update completed successfully")
+      elsif statuses[job_id][:status] == "Failed"
+        raise("Firmware update failed in the lifecycle controller.  Please refer to LifeCycle job logs")
+      elsif statuses[job_id][:status] == "Downloaded"
+        logger.debug("Firmware downloaded to idrac")
+      end
+      statuses
+    end
+
+    # Used to schedule the reboot job queue
+    #
+    # @param reboot_firmwares [Hash] contains list of firmwares which requires reboot
+    # @option reboot_firmwares [String] :instance_id contains job_id
+    # @option reboot_firmwares [String] :uri_path contains NFS mount path
+    # @param force_restart [Boolean]
+    # @param wsman [Object]
+    # @return [void]
+    def schedule_reboot_job_queue(reboot_firmwares, force_restart, wsman)
+      unless reboot_firmwares.empty?
+        reboot_id = nil
+        if force_restart
+          logger.debug("Creating Reboot Job")
+          resp = wsman.create_reboot_job(:reboot_job_type => :power_cycle, :timeout => 5 * 60)
+          reboot_id = resp[:reboot_job_id]
+          logger.debug("Reboot Job scheduled successfully")
+        end
+        logger.debug("Setting the Job queue")
+
+        if reboot_id
+          wsman.setup_job_queue(:job_array => reboot_id, :start_time_interval => "TIME_NOW")
+        end
+
+        reboot_firmwares.keys.each do |job_id|
+          wsman.setup_job_queue(:job_array => job_id, :start_time_interval => "TIME_NOW")
+        end
+
+        if force_restart
+          reboot_status = "new"
+          until reboot_status == "Reboot Completed"
+            sleep 30
+            lc_job = wsman.get_lc_job(reboot_id)
+            reboot_status = lc_job[:job_status]
+            logger.debug("Reboot Status: #{reboot_status}")
+          end
+        end
+      end
+    end
+
+    # Used to create an XML file:
+    #
+    # @param instance_id [String] job_id
+    # @return [void]
+    # Path specifies moount path
+    def create_xml_config_file(instance_id, path)
+      template = <<-EOF
+<p:InstallFromURI_INPUT xmlns:p="http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_SoftwareInstallationService">
+<p:URI><%= path %></p:URI>
+<p:Target xmlns:a="http://schemas.xmlsoap.org/ws/2004/08/addressing" xmlns:w="http://schemas.dmtf.org/wbem/wsman/1/wsman.xsd">
+<a:Address>http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous</a:Address>
+<a:ReferenceParameters>
+<w:ResourceURI>http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_SoftwareIdentity</w:ResourceURI>
+<w:SelectorSet>
+<w:Selector Name="InstanceID"><%= instance_id %></w:Selector>
+</w:SelectorSet> </a:ReferenceParameters> </p:Target> </p:InstallFromURI_INPUT>
+      EOF
+      xmlout = ERB.new(template)
+      temp_file = Tempfile.new("xml_config")
+      temp_file.write(xmlout.result(binding))
+      temp_file.close
+      temp_file
+    end
+  end
+end
--- a/lib/asm/transport/racadm.rb
+++ b/lib/asm/transport/racadm.rb
@@ -1,0 +1,39 @@
+require "net/ssh"
+
+module ASM
+  class Transport
+    class IdracResetError < StandardError; end
+    class Racadm
+      attr_reader :endpoint, :logger
+
+      def initialize(endpoint, log)
+        @logger = log
+        @endpoint = endpoint
+      end
+
+      # Used to reset iDRAC
+      #
+      # @return [void]
+      # @raise [StandardError] the resulting would unable to establish ssh connection or racadm command fails
+      def reset_idrac
+        logger.debug("Resetting iDrac...")
+        Net::SSH.start(@endpoint[:host],
+                       @endpoint[:user],
+                       :password => @endpoint[:pass],
+                       :paranoid => Net::SSH::Verifiers::Null.new,
+                       :global_known_hosts_file => "/dev/null") do |ssh|
+          ssh.exec!("racadm racreset soft") do |_, stream, data|
+            logger.debug(data)
+
+            if data.include? "Could not chdir to home directory"
+              logger.debug("Warning for message is %s" % data)
+            elsif stream == :stderr
+              raise IdracResetError
+            end
+          end
+          ssh.close
+        end
+      end
+    end
+  end
+end
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -538,6 +538,17 @@ module ASM
                     :return_value => "4096")
     end
 
+    # Parse the specified XML generated schema
+    #
+    # @params options [Hash] the options to install specified config file path
+    # @example
+    #      :input_file => "/tmp/xml_parsed_path"
+    # @return [Hash]
+    # @raise [ResponseError] if command fails
+    def install_from_uri(options={})
+      client.invoke("InstallFromURI", SOFTWARE_INSTALLATION_SERVICE, :input_file => options[:input_file])
+    end
+
     # Set up a job queue with jobs to execute
     #
     # This method is used for creating a job queue that shall contain one or more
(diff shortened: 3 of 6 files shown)


lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,151 @@
...
+    MAX_WAIT_SECONDS = 1800
+    attr_reader :endpoint, :logger, :wsman
+
+    def initialize(cred, options= {})
+      @endpoint = cred
+      @logger = options[:logger]
+      @wsman = ASM::WsMan.new(@endpoint, options)
+    end
+
+    def update_idrac_firmware(firmware_list, force_restart, logger)
```
ripienaar: this method is over 100 lines, please make small ones where each method have a single responsibily and then write tests

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,28 @@
+require 'net/ssh'
+
+module ASM
+  class IdracResetError < Exception;
+  end
+  class Racadm
```
ripienaar: this is in `asm/transport/racadm` but the class is `ASM::Racadm`, that's not right, these things have to align

lib/asm/util.rb
```diff
@@ -242,6 +245,64 @@ def self.run_command(cmd, *args)
...
+      #        "server_firmware"=>
+      #                "[{\"instance_id\":\"DCIM:INSTALLED#701__NIC.Embedded.1-1-1\",\"uri_path\":\"/var/nfs/firmware/ff8080815746abab0157481ecea1000f/FOLDER03287319M/3/Network_Firmware_0MT4K_WN64_7.10.64.EXE\"}]"}}}
+    # resource_hash: Array of instances with nfs and uri paths
+    # For instance
+     # [{"instance_id"=>"DCIM:INSTALLED#701__NIC.Embedded.1-1-1",
+     #   "uri_path"=>"nfs://172.25.5.100/FOLDER03287319M/3/Network_Firmware_0MT4K_WN64_7.10.64.EXE;mountpoint=/var/nfs/firmware/ff808081578bd88601578d525d4e004e"},
+     #  {"instance_id"=>"DCIM:INSTALLED#iDRAC.Embedded.1-1#IDRACinfo",
+     #   "component_id"=>"25227",
+     #   "uri_path"=> nfs://172.25.5.100/FOLDER03526343M/2/iDRAC-with-Lifecycle-Controller_Firmware_JHF76_WN64_2.30.30.30_A00.EXE;mountpoint=/var/nfs/firmware/ff808081578bd88601578d525d4e004e"}]
+    # logger: logger console
```
ripienaar: please use YARD format and correctly indent things

lib/asm/util.rb
```diff
@@ -242,6 +245,64 @@ def self.run_command(cmd, *args)
...
+      raise (ASM::Error, "Received empty resources to update the firmware on server") if resource_hash.nil?
+
+      device_conf = ASM::DeviceManagement.parse_device_config(cert_name)
+      options = {:host=>device_conf["host"], :user=>device_conf["user"], :password=>device_conf["password"] }
+      wsman = ASM::WsMan.new(options, :logger => logger)
+      wsman.clear_job_queue_retry(options, logger)
+      force_restart = config["asm::server_update"][cert_name]["force_restart"]
+      pre =[]
+      main = []
+      resource_hash.each do |firmware|
```
ripienaar: in long methods - or really any method - where there are many blocks and many stages of flow adding empty lines around blocks helps with readability.

Your method tells a story to those reading it, stories without paragraphs all squashed up together is hard to read.

lib/asm/util.rb
```diff
@@ -242,6 +245,64 @@ def self.run_command(cmd, *args)
...
+      end
+      firmware_obj = ASM::Firmware.new(options, :logger=> logger)
+      if pre.size > 0
+        logger.debug("LC Update required,installing first")
+          firmware_obj.update_idrac_firmware(pre,force_restart, logger)
+      end
+      firmware_obj.update_idrac_firmware(main,force_restart, logger)
+      # After updating Ensure LC is up and in good state before exiting
+      status = wsman.remote_services_api_status
+      if status.to_i == 0
```
ripienaar: best to use `Integer(status)` since `"a".to_i` will give `0` too.

lib/asm/util.rb
```diff
@@ -242,6 +245,64 @@ def self.run_command(cmd, *args)
...
+        lc_ready = true
+        # break
+      else
+        logger.debug("LC status is busy: status code #{status}. Waiting ...")
+        sleep 60
```
ripienaar: why sleep here and then do nothing? shouldn't we like check it again or something? I imagine you wanted to check again with some retry counter and then set lc_ready after timeout, now when we enter this bit here lc_ready is never set and your exception below will always raise an exception because your `unless lc_ready` is on a unset variable.

lib/asm/wsman.rb
```diff
@@ -574,6 +579,49 @@ def delete_job_queue(params={})
                     :return_value => "0")
     end
 
+    #Clearing the job_queue with retry
```
ripienaar: write YARD style comments please

lib/asm/util.rb
```diff
@@ -242,6 +245,64 @@ def self.run_command(cmd, *args)
...
+      #                "[{\"instance_id\":\"DCIM:INSTALLED#701__NIC.Embedded.1-1-1\",\"uri_path\":\"/var/nfs/firmware/ff8080815746abab0157481ecea1000f/FOLDER03287319M/3/Network_Firmware_0MT4K_WN64_7.10.64.EXE\"}]"}}}
+    # resource_hash: Array of instances with nfs and uri paths
+    # For instance
+     # [{"instance_id"=>"DCIM:INSTALLED#701__NIC.Embedded.1-1-1",
+     #   "uri_path"=>"nfs://172.25.5.100/FOLDER03287319M/3/Network_Firmware_0MT4K_WN64_7.10.64.EXE;mountpoint=/var/nfs/firmware/ff808081578bd88601578d525d4e004e"},
+     #  {"instance_id"=>"DCIM:INSTALLED#iDRAC.Embedded.1-1#IDRACinfo",
+     #   "component_id"=>"25227",
+     #   "uri_path"=> nfs://172.25.5.100/FOLDER03526343M/2/iDRAC-with-Lifecycle-Controller_Firmware_JHF76_WN64_2.30.30.30_A00.EXE;mountpoint=/var/nfs/firmware/ff808081578bd88601578d525d4e004e"}]
+    # logger: logger console
+    def self.idrac_fw_install_fromuri(cert_name, config, resource_hash, logger)
```
ripienaar: I'd make this `....from_uri`

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,151 @@
...
+module ASM
+  class Firmware
+    IDRAC_ID = 25227
+    LC_ID = 28897
+    UEFI_DIAGNOSTICS_ID = 25806
+    DRIVER_PACK = 18981
+    OS_COLLECTOR = 101734
+
+    # Component ids that do not require a reboot (DIRECT UPDATES)
+    NO_REBOOT_COMPONENT_IDS = [IDRAC_ID, LC_ID, UEFI_DIAGNOSTICS_ID, DRIVER_PACK, OS_COLLECTOR]
```
kaushalgala: I don't think this is a good way to determine updates do not require reboot.

I thought there was a way to figure out if a component update via idrac requires reboot.  @gavin-scott may know it.

lib/asm/idrac_fw_update.rb
kaushalgala: yes, we should break it down into several small methods, and this method should look like an orchestration code

kaushalgala: provide yard doc for the method update_idrac_firmware, and any new methods you break this into. 

Also not sure why we pass logger, when we already define logger instance variable in line 23 ?

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,151 @@
...
+        if response[:return_value] == '4096'
+          job_id = resp[:job]
+          logger.debug("InstallFromURI started")
+          logger.debug("JOB_ID: #{job_id}")
+        else
+          logger.debug("Error installing From URI config: #{config_file.read}")
+          raise(ASM::Error, "Problem running InstallFromURI: #{resp[:message]}")
+        end
+        raise("Failed to initiate the firmware job for #{fw}") unless job_id
+        raise("Duplicate job id #{job_id} for fimrware #{fw}: #{statuses[job_id]}") if statuses[job_id]
```
kaushalgala: fix spelling for "firmware"

kaushalgala: I guess in new code we write, we follow string substitution as "%s" % var instead of "#{var}":

Can you change to:
"Failed to initiate the firmware job for %s" % fw
"Duplicate job id %s for fimrware %s: %s" % [job_id, fw, statuses[job_id] ]

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,151 @@
...
+
+        statuses[job_id] = {:job_id => job_id, :status => 'new', :firmware => fw, :start_time => Time.now}
+        until statuses[job_id][:status] =~ /Downloaded|Completed|Failed/
+          sleep 30
+          begin
+            statuses[job_id][:status] = wsman.checkjobstatus(endpoint, job_id, logger)
+          rescue Exception => e
+            statuses[job_id][:status] = 'TemporaryFailure'
```
kaushalgala: use double quote instead of single quote for strings

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,151 @@
...
+            logger.warn(" Look up job status #{job_id} failed: #{e}")
+          end
+          logger.debug("Job Status: #{statuses[job_id][:status]}")
+          if Time.now - statuses[job_id][:start_time] > MAX_WAIT_SECONDS
+            logger.warn("Timed out waiting for firmware job #{job_id} to complete")
+            statuses[job_id][:status] = 'Failed'
```
kaushalgala: same comment here, use double quote for strings

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,151 @@
...
+        raise("Duplicate job id #{job_id} for fimrware #{fw}: #{statuses[job_id]}") if statuses[job_id]
+
+        statuses[job_id] = {:job_id => job_id, :status => 'new', :firmware => fw, :start_time => Time.now}
+        until statuses[job_id][:status] =~ /Downloaded|Completed|Failed/
+          sleep 30
+          begin
+            statuses[job_id][:status] = wsman.checkjobstatus(endpoint, job_id, logger)
+          rescue Exception => e
+            statuses[job_id][:status] = 'TemporaryFailure'
+            logger.warn(" Look up job status #{job_id} failed: #{e}")
```
kaushalgala: remove space before "Look"

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def initialize(endpoint, log)
+      @logger = log
+      @endpoint = endpoint
+    end
+
+    def reset_idrac
+      logger.info("Resetting IDrac...")
+      ssh = Net::SSH.start(@endpoint[:host], @endpoint[:user], :password => @endpoint[:pass])
+      data = ssh.exec!("racadm racerset soft")
```
kaushalgala: this command is wrong, must be racreset , not racerset :)

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,28 @@
...
+  class Racadm
+    attr_reader :endpoint, :logger
+
+    def initialize(endpoint, log)
+      @logger = log
+      @endpoint = endpoint
+    end
+
+    def reset_idrac
+      logger.info("Resetting IDrac...")
```
kaushalgala: IDrac => iDRAC

lib/asm/util.rb
```diff
@@ -242,6 +245,64 @@ def self.run_command(cmd, *args)
...
+          logger.debug(firmware)
+           if [LC_ID , IDRAC_ID].include? firmware['component_id'].to_i
+             pre << firmware
+           else
+              main << firmware
+           end
+      end
+      firmware_obj = ASM::Firmware.new(options, :logger=> logger)
+      if pre.size > 0
+        logger.debug("LC Update required,installing first")
```
kaushalgala: add space between comma and new word

prem548:
@kaushalgala I think , I addressed most of comments suggested. Could you review it. Please let me know anything necessary.

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+module ASM
+  class Firmware
+    IDRAC_ID = 252_27
+    LC_ID = 288_97
+    UEFI_DIAGNOSTICS_ID = 258_06
+    DRIVER_PACK = 189_81
+    OS_COLLECTOR = 101_734
+
+    # Component ids that do not require a reboot (DIRECT UPDATES)
+    NO_REBOOT_COMPONENT_IDS = [IDRAC_ID, LC_ID, UEFI_DIAGNOSTICS_ID, DRIVER_PACK, OS_COLLECTOR].freeze
```
kaushalgala: let's discuss with @gavin-scott , if there is better way to tell an update component needs reboot

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+      @endpoint = cred
+      @logger = options[:logger]
+      @wsman = ASM::WsMan.new(@endpoint, options)
+    end
+
+    # Used to update the firmware on each NIC slot
+    #
+    # @param firmware_list Array[<Hash>] Each instance with mount and nfs path
+    # @param force_restart [Boolean]
+    def update_idrac_firmware(firmware_list, force_restart)
```
kaushalgala: Add to yard doc about exception.  I can see this function raises a StandardError

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+        logger.info("Failed firmware jobs: #{failures}")
+        raise("Firmware update failed in the lifecycle controller. Please refer to LifeCycle job logs")
+      end
+    end
+
+    # Used to install specified firmware instance on NIC slot
+    #
+    # @param firmware [Hash]
+    # @return [String] created JOB_ID
+    def gets_install_uri_job(firmware)
```
kaushalgala: same comment on adding yard doc on exception raised

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+        logger.debug("Error installing From URI config: #{config_file.read}")
+        raise(ASM::Error, "Problem running InstallFromURI: #{resp[:message]}")
+      end
+      job_id
+    end
+
+    # Helper method used to block till firmware status completed
+    #
+    # @param job_id [String]
+    # @return [Hash]
```
kaushalgala: document what is the return hash and possible options on returned hash

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+        raise("Firmware update failed in the lifecycle controller.  Please refer to LifeCycle job logs")
+      elsif statuses[job_id][:status] == "Downloaded"
+        logger.debug("Firmware downloaded to idrac")
+      end
+      statuses
+    end
+
+    # Used to schedule the reboot job queue
+    #
+    # @param reboot_firmwares [Hash]
```
kaushalgala: document the reboot_firmwares variable, and its possible options.

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+          statuses[job_id][:status] = "TemporaryFailure"
+          logger.warn("Look up job status #{job_id} failed: #{e}")
+        end
+        logger.debug("Job Status: #{statuses[job_id][:status]}")
+        if Time.now - statuses[job_id][:start_time] > MAX_WAIT_SECONDS
+          logger.warn("Timed out waiting for firmware job #{job_id} to complete")
+          statuses[job_id][:status] = "Failed"
+        end
+      end
+      if statuses[job_id][:status] == "Completed"
```
kaushalgala: add space above this line for readability

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+      unless reboot_firmwares.empty?
+        reboot_id = nil
+        if force_restart
+          logger.debug("Creating Reboot Job")
+          resp = wsman.create_reboot_job(:reboot_job_type => :power_cycle, :timeout => 5 * 60)
+          reboot_id = resp[:reboot_job_id]
+          logger.debug("Reboot Job scheduled successfully")
+        end
+        logger.debug("Setting the Job queue")
+        if reboot_id
```
kaushalgala: add space above.

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,188 @@
...
+          wsman.setup_job_queue(:job_array => reboot_id, :start_time_interval => "TIME_NOW")
+        end
+        reboot_firmwares.keys.each do |job_id|
+          wsman.setup_job_queue(:job_array => job_id, :start_time_interval => "TIME_NOW")
+        end
+        if force_restart
```
kaushalgala: add space above.

lib/asm/wsman.rb
```diff
@@ -574,6 +585,62 @@ def delete_job_queue(params={})
...
+        retry
+      end
+    end
+
+    # check the running Job status by providing the job_id
+    #
+    # @param endpoint [Hash] contains the credential to establish a session
+    # @param instance_id [String] JOB_ID or RID_ID
+    # @param logger [Logger]
+    # @return [Hash]
```
kaushalgala: document detail on return value, including possible hash values.. specify example if applicable

lib/asm/idrac_fw_update.rb
gavin-scott: Looks like the info is in the catalog:

```
  <SoftwareComponent schemaVersion="1.0" packageID="8R65X" releaseID="8R65X" hashMD5="C957B47625BBBF2142CDBC386617DA97" path="FOLDER01996491M/1/SAS-Drive_Firmware_8R65X_WN64_K772_A01.EXE" dateTime="2014-02-04T16:41:48+00:00" releaseDate="February 04, 2014" vendorVersion="K772" dellVersion="A01" packageType="LW64" rebootRequired="false">
```

It would be good to use that info for which components require reboot rather than hard-coding it here. The example above is for hard drive firmware which doesn't require a reboot, but because of our hard-coded list above we would reboot anyway.

This is what the current code in the dell-idrac module does however, and it would require a bit of a refactor to get the catalog information passed in from asm-core so I'm going to just file this as an enhancement jira item.

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,206 @@
...
+      firmware_list.each do |fw|
+        logger.debug(fw)
+        job_id = gets_install_uri_job(fw)
+        raise("Failed to initiate the firmware job for %s" % fw) unless job_id
+        raise("Duplicate job id %s for firmware %s: %s" % [job_id, fw, statuses[job_id]]) if statuses[job_id]
+        statuses = block_until_downloaded(job_id, fw)
+      end
+
+      statuses.each do |_, v|
+        if NO_REBOOT_COMPONENT_IDS.include?(v[:firmware]["component_id"].to_i)
```
ripienaar: please do not use `to_i` use `Integer(thing)` instead, `"a".to_i` is `0`, we dont want to let corrupt input data just silently change behaviour so `Integer(thing)` will raise a error in the case of `Integer("a")`

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,206 @@
...
+
+    def initialize(cred, options={})
+      @endpoint = cred
+      @logger = options[:logger]
+      @wsman = ASM::WsMan.new(@endpoint, options)
+    end
+
+    # Used to update the firmware on each NIC slot
+    #
+    # @param firmware_list Array[<Hash>] Each instance with mount and nfs path
```
ripienaar: `@param firmware_list [Array<Hash>] ....`

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,206 @@
...
+      else
+        logger.info("Failed firmware jobs: #{failures}")
+        raise("Firmware update failed in the lifecycle controller. Please refer to LifeCycle job logs")
+      end
+    end
+
+    # Used to install specified firmware instance on NIC slot
+    #
+    # @param firmware [Hash]
+    # @options [String] :instance_id specifies the particular NIC slot
```
ripienaar: these would be `@option firmware [String] :instance_id foo foo foo`

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,206 @@
...
+        raise(ASM::Error, "Problem running InstallFromURI: #{resp[:message]}")
+      end
+      job_id
+    end
+
+    # Helper method used to block till firmware status completed
+    #
+    # @param job_id [String]
+    # @return [Hash] statuses
+    # * :job_id [Hash] key contains the following schema
```
ripienaar:
```
# @param job_id [Hash]
# @option job_id [String] :job_id created job id
# @option job_id ["now", "Failed", "TemporaryFailure", "Downloaded", "Completed"] :status job status
```

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,206 @@
...
+    # @options [String] :job_id created job_id
+    # @options [String] :status defines the status of job possible values "new" or "Failed" or "TemporaryFailure" or "Downloaded" or "Completed"
+    # @raise [StandardError]  if the checkjobstatus command fails
+    # @raise [Error] if the firmware job status was failed
+    def block_until_downloaded(job_id, firmware)
+      statuses = {}
+      statuses[job_id] = {:job_id => job_id, :status => "new", :firmware => firmware, :start_time => Time.now}
```
ripienaar: combine these two into 

```
statusses = {
  job_id => {
    ......
  }
}
```

lib/asm/idrac_fw_update.rb
```diff
@@ -0,0 +1,206 @@
...
+        logger.debug("Firmware downloaded to idrac")
+      end
+      statuses
+    end
+
+    # Used to schedule the reboot job queue
+    #
+    # @param reboot_firmwares [Hash] contains list of firmwares which requires reboot
+    # @options [String] :instance_id contains job_id
+    # @options [String] :uri_path contains NFS mount path
```
ripienaar: as above examples

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,29 @@
...
+    class IdracResetError < Exception; end
+    class Racadm
+      attr_reader :endpoint, :logger
+
+      def initialize(endpoint, log)
+        @logger = log
+        @endpoint = endpoint
+      end
+
+      def reset_idrac
```
ripienaar: docs please

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,29 @@
+require "net/ssh"
+
+module ASM
+  class Transport
+    class IdracResetError < Exception; end
```
ripienaar: You'd inherit from `StandardError` instead of `Exception`, Exception is for really terrible things like out of memory, standard error is for user flow/feedback stuff

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,29 @@
...
+        @endpoint = endpoint
+      end
+
+      def reset_idrac
+        logger.debug("Resetting iDrac...")
+        ssh = Net::SSH.start(@endpoint[:host], @endpoint[:user], :password => @endpoint[:pass])
+        data = ssh.exec!("racadm racreset soft")
+        ssh.close
+
+        logger.debug(data)
```
ripienaar: when logging, even at debug` please make sure that if someone sees your log line they can figure out what its about, just dumping `data` might not have enough context to tell someone what he is looking at

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,29 @@
...
+        if data.include? "Could not chdir to home directory"
```
ripienaar: hmm, this seems fragile and might fail randomly if the ssh library start giving other errors, is there no error code or class or something? I think you'd rather use `if data.match("Could not....")` than `include?` here

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,29 @@
...
+          logger.debug "Warning for message - #{data}"
```
ripienaar: same comment about log message context as above - also use printf style please

lib/asm/util.rb
```diff
@@ -242,6 +245,77 @@ def self.run_command(cmd, *args)
       result
     end
 
+    # Used to update IDRAC_FW
```
ripienaar: new line below

lib/asm/util.rb
```diff
@@ -242,6 +245,77 @@ def self.run_command(cmd, *args)
...
+    # Params:
```
ripienaar: not needed, please remove

lib/asm/util.rb
```diff
@@ -242,6 +245,77 @@ def self.run_command(cmd, *args)
...
+    # @param cert_name [String] specifies IDRAC_certname
+    # @param config [HASH] schema of resources used to update the firmware
```
ripienaar: `Hash`

lib/asm/util.rb
```diff
@@ -242,6 +245,77 @@ def self.run_command(cmd, *args)
...
+    #      [{"instance_id" => "DCIM:INSTALLED#701__NIC.Embedded.1-1-1",
+    #        "uri_path" => "nfs://172.25.5.100/FOLDER03287319M/3/Network_Firmware_0MT4K_WN64_7.10.64.EXE;mountpoint=/var/nfs/firmware/ff808081578bd88601578d525d4e004e"
+    #       },
+    #       {"instance_id" => "DCIM:INSTALLED#iDRAC.Embedded.1-1#IDRACinfo",
+    #        "component_id" => "25227",
+    #        "uri_path" => "nfs://172.25.5.100/FOLDER03526343M/2/iDRAC-with-Lifecycle-Controller_Firmware_JHF76_WN64_2.30.30.30_A00.EXE;
+    #                       mountpoint="/var/nfs/firmware/ff808081578bd88601578d525d4e004e"
+    #       }]
+    # @param logger [logger]
+    # @return [void]
```
ripienaar: document the exceptions too please

lib/asm/wsman.rb
```diff
@@ -574,6 +586,64 @@ def delete_job_queue(params={})
...
+    # Clearing the job_queue with retry
+    #
+    # make sure lc ready after clearing the job queue other wise reset the IDRAC
+    # @param endpoint [Hash] contains the credential to establish a session
+    # @param logger [Logger] console
+    # @return [Void] make sure lc ready other wise reset the IDRAC
+    # @raise [StandardError] if the command delete_job_queue operation or lc_status was not ready after job queue deletion
+    def clear_job_queue_retry(endpoint, logger)
+      attempts = 0
+      begin
```
ripienaar: some spacing around begine, if, end etc

lib/asm/wsman.rb
```diff
@@ -574,6 +586,64 @@ def delete_job_queue(params={})
...
+      job_status, job_message, message_id = ASM::WsMan.invoke(endpoint, "get", schema,
+                                                              :logger => logger,
+                                                              :selector => ["//n1:JobStatus", "//n1:Message", "//n1:MessageID"])
+
+      if message_id =~ /SYS051|LC068/i
+        message_id
+      elsif job_status =~ /completed with errors/i || job_message =~ /completed with errors/i
+        "Failed"
+      elsif job_status.nil? || job_status.empty?
+        raise "Job ID not created"
```
ripienaar: you could probably give a bit more context here in this message, like what was the status for example?

lib/asm/idrac_fw_update.rb
kaushalgala: yeah, good point - use `Integer(v[:firmware]["component_id"])`

prem548:
@ripienaar First of all Wow. Nice comments and suggestions. I had little concern on writing specs. Every changes I made includes wsman commands. I have to stub it check it with whether its been invoked or not  . Some cases I have to create fixtures and test it. Do you want me do All of that.  I will test end to end with those changes will it be ok? Other than  that I would be happy to make changes.

lib/asm/transport/racadm.rb
prem548: Its not printing anything about cred. the Dumping data includes: 

RAC reset operation initiated successfully. It may take a few
 minutes for the RAC to come online again.

Thats the reason why I just dumped the data variable into logger

lib/asm/idrac_fw_update.rb
gavin-scott: Perhaps in future we could enhance this to read the "reboot" metadata from the catalog XML. However right now the catalog XML isn't passed through to this stage. I think this is ok for now.

lib/asm/transport/racadm.rb
```diff
@@ -0,0 +1,29 @@
...
+        elsif data == :stderr
```
gavin-scott: I don't think this is a valid check. From the API docs if `ssh.exec!` does not have a block the output will always be a string: http://net-ssh.github.io/net-ssh/Net/SSH/Connection/Session.html#method-i-exec-21

lib/asm/transport/racadm.rb
gavin-scott: Since we're running this over ssh and the the iDrac has it's own custom shell (not bash or anything) I can't see any other way to get error information other than parsing the output. This is supposed to be an exception case where we allow this error message but fail on all others. I think this needs to be reworked to pass a block to `exec!` like this:

https://github.com/dell-asm/dell-idrac/blame/79fd5fb2f4d7510a636a36103790c1a9ce09f676/lib/puppet/idrac/util.rb#L169-L174

ripienaar:
@prem548 yes we need tests especially if they are hard to write - because that tends to tell you the code might be over complex

For the wsman commands you can probably create a method to run wsman for you based on some arguments and  then testing becomes easy because you check you receive the right arguments. Lots of examples I bet in the wsman classes Gavin made

Yes test fixtures would be great to help you test this too especially those big methods. But small single purpose methods with small bits of input data helps there too. It seems like making 10 method a is pointless when 1 doesn't read too hard - but it really pays off in being able to test the code

lib/asm/wsman.rb
```diff
@@ -574,6 +586,64 @@ def delete_job_queue(params={})
...
+      begin
+        attempts += 1
+        logger.debug("Clearing the Job Queue...")
+        resp = delete_job_queue(:job_id => "JID_CLEARALL")
+        if resp[:return_value] == "0"
+          logger.debug("Jobs in job queue was cleared successfully...")
+        else
+          raise(ASM::Error, "Unable to clear all the job queue")
+        end
+        value = remote_services_api_status
```
gavin-scott: I would suggest using `poll_for_lc_ready` here to make sure LC is ready.

lib/asm/wsman.rb
```diff
@@ -574,6 +586,64 @@ def delete_job_queue(params={})
...
+        if value[:lcstatus] == "0"
+          logger.debug("LC controllers Remote services are ready ...")
+        end
+      rescue StandardError => e
+        logger.debug("Job queue cannot be cleared.") if attempts > 1
```
gavin-scott: We should have some kind of overall timeout here, so that we can't e.g. get to attemps == 50000

lib/asm/wsman.rb
```diff
@@ -574,6 +586,64 @@ def delete_job_queue(params={})
...
+    end
+
+    # check the running Job status by providing the job_id
+    #
+    # @param endpoint [Hash] contains the credential to establish a session
+    # @param instance_id [String] JOB_ID or RID_ID
+    # @param logger [Logger]
+    # @return [String] returns message_id or job_status or "Failed"
+    # @raise [Error]  job_id was not created
+    def checkjobstatus(endpoint, instance_id, logger)
```
gavin-scott: I think this whole thing can go away and use `get_lc_job`:

https://github.com/dell-asm/dell-asm-util/blob/af992b80efb7288ee25da7152eeb3d903fe97c88/lib/asm/wsman.rb#L1138

gavin-scott:
Re unit tests -- yes, please add unit tests for at least any of the new `ASM::WsMan` methods. For the existing firmware update logic if there were tests in the dell-idrac module port those over. If not it may be ok to leave that out for now so we can get these changes in.

lib/asm/idrac_fw_update.rb
ripienaar: @prem548 ok, yeah I guess in this case that would be safe to do, happy to leave as is.

gavin-scott: I assume every firmware has a component_id, is that not the case? If every firmware has a component_id then your example is exactly the point -- we'll fail on invalid input.

prem548: @gavin-scott sometimes we do get firmwares that do not have a component_id

lib/asm/firmware.rb
```diff
@@ -0,0 +1,330 @@
...
+    #       "port"=>nil,
+    #       "path"=>"/asm/bin/idrac-discovery.rb",
+    #       "scheme"=>"script",
+    #       "arguments"=>{"credential_id"=>"ff80808157bfd05a0157bfd13392000d"},
+    #       "user"=>"username",
+    #       "enc_password"=>nil,
+    #       "password"=> "1234"
+    #      }
+    #
+    # @param logger [logger]
```
gavin-scott: Capital L `Logger`

lib/asm/firmware.rb
```diff
@@ -0,0 +1,330 @@
...
+        end
+      end
+    end
+
+    # Used to update the firmware on each NIC slot
+    #
+    # @param firmware_list [Array[<Hash>]] Each instance with mount and nfs path
+    # @param force_restart [Boolean]
+    # @param wsman [Object]
+    # @raise [StandardError] if firmware gets_install_uri_job was not able to create job_id to receive or created duplicate job_ids or firmware update fails
```
gavin-scott: Need `@return [void]` before this line.

lib/asm/firmware.rb
```diff
@@ -0,0 +1,330 @@
...
+      failures = statuses.values.find_all { |v| v[:status] =~ /Failed|InternalTimeout/ }
+      if failures.empty?
+        logger.debug("Firmware update completed successfully")
+      else
+        logger.info("Failed firmware jobs: #{failures}")
+        raise("Firmware update failed in the lifecycle controller. Please refer to LifeCycle job logs")
+      end
+    end
+
+    # Used to install specified firmware instance on NIC slot
```
gavin-scott: This isn't specific to NICs is it?

lib/asm/firmware.rb
```diff
@@ -0,0 +1,330 @@
...
+    # @param firmware [Hash]
+    # @param wsman [Object]
+    # @option firmware [String] :instance_id defines specific NIC slot
+    # @option firmware [String] :uri_path defines specific mount share path
+    # @option firmware [String] :component_id by default component id is "25227"
+    # @return statuses [Hash]
+    # @option statuses [String] :job_id created job_id
+    # @option statuses ["new", "Failed", "TemporaryFailure", "Downloaded", "Completed"] :status statuses
+    # @option statuses [String] :firmware defines specific NIC slot firmware
+    # @raise [StandardError]  if the checkjobstatus command fails
```
gavin-scott: Need `@return [void]`

lib/asm/firmware.rb
prem548: It will receive on firmware local variable 
{"instance_id" => "DCIM:INSTALLED#701__NIC.Slot.2-2-1", "uri-path" => "path"} first then followed by {"instance_id" => "DCIM:INSTALLED#701__NIC.Embedded.1-1-1" "uri-path"=> "path"} ....... and so on ..

prem548:
I updated the comments suggested please let me know if any other changes necessary. @gavin-scott

lib/asm/firmware.rb
gavin-scott: The `@return` should go before the `@raise`. Please double-check this on all of the YARD docs in this PR and make sure they all conform, then ping me when ready for me to re-review.

lib/asm/firmware.rb
```diff
@@ -0,0 +1,332 @@
...
+      end
+      job_id
+    end
+
+    # Helper method used to block till firmware status completed
+    #
+    # @param job_id [String]
+    # @param firmware [Hash]
+    # @param wsman [Object]
+    # @option firmware [String] :instance_id defines specific NIC slot
```
gavin-scott: It's not specific to NIC slot, it's the general firmware instance ID

lib/asm/firmware.rb
```diff
@@ -0,0 +1,333 @@
...
+          transport.reset_idrac
+          sleep 60
+          retry
+        else
+          raise("Unable to find the LC status after clearing job queue")
+        end
+      end
+    end
+
+    # Used to update the IDRAC firmware
```
gavin-scott: Please change to "Used to update server firmware via the iDrac"

PR #76, opened 2016-10-10
ASM-7905 Utility method - wsman identify command

kaushalgala:
wsman identify command provides quick information about a server, and
can be used for scenarios like waiting on valid IDRAC server when IP
changes.

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1483,6 +1483,36 @@ module ASM
       end
     end
 
+    # Execute identify command
+    #
+    # executes identify command on the server, and returns raw output
+    #
+    # @param timeout [FixNum] command timeout in seconds
+    # @return [Hash] nil if there was a command timeout, otherwise output result
+    # @example response
+    #  {
+    #    :protocol_version=>"http://schemas.dmtf.org/wbem/wsman/1/wsman.xsd",
+    #    :product_vendor=>"Dell, Inc.",
+    #    :product_version=>"iDRAC : System Type = 12G Monolithic : LC Version = 2.30.30.30 : Version = 2.30.30.30",
+    #    :smashversion=>"2.0.0",
+    #    :product_name=>"iDRAC",
+    #    :system_generation=>"12G Monolithic",
+    #    :firmware_version=>"2.30.30.30",
+    #    :lifecycle_controller_version=>"2.30.30.30",
+    #    :security_profiles=>"HTTP_TLS_1HTTP_TLS_2"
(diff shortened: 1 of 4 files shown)


lib/asm/wsman.rb
```diff
@@ -1483,6 +1483,29 @@ def self.wait_for_lc_ready(endpoint, logger=nil, attempts=0, max_attempts=30)
...
+    # Execute wsman identify command
+    #
+    # executes identify command on the server, and returns raw output
+    #
+    # @param ip [String] IP address of the server
+    # @param username [String] username credential for connecting to the server
+    # @param password [String] password credential for connecting to the server
+    # @param timeout [Integer] maximum seconds before server command times out
+    # @return [String] nil if stderr is not empty, otherwise raw stdout response
+    def self.exec_identify(ip, username, password, timeout=30)
```
gavin-scott: I would call this just `identify`

lib/asm/wsman.rb
```diff
@@ -1483,6 +1483,29 @@ def self.wait_for_lc_ready(endpoint, logger=nil, attempts=0, max_attempts=30)
...
+      err = ""
+      cmd = ["wsman", "identify", "-h", ip, "-P", "443", "-u", username, "-p", password,
+             "-c", "dummy.cert", "-y", "basic", "-V", "-v", "--transport-timeout=%d" % timeout]
+      Open3.popen3(*cmd) do |stdin, stdout, stderr, _wait_thr|
+        stdin.close
+        output = stdout.read
+        err = stderr.read
+      end
+      # return output when stderr is empty, else return nil
+      output if err.empty?
```
gavin-scott: We should use the underlying `ASM::WsMan::Client` here. I think the required changes are:

1. Add an `ASM::WsMan::Client#identify` method similar to enumerate: https://github.com/dell-asm/dell-asm-util/blob/master/lib/asm/wsman/client.rb#L195
2. Make it able to handle identify by adding that as a base option here like enumerate and get: https://github.com/dell-asm/dell-asm-util/blob/dfc33cb738f751b3e307f349c8f17f3f611d87c4/lib/asm/wsman/client.rb#L68
3. Since `identify` doesn't take a schema you'll need to ignore that in part 2 above

Please add unit tests for all the methods, I try to keep a high test coverage on the ASM::WsMan stuff.

kaushalgala:
@gavin-scott made changes as you suggested, and added unit tests.

Please review the changes.

gavin-scott:
Nice!

PR #78, opened 2016-11-02
ASM-7884, ASM-8182 Firmware updates failing

logankimmel:
Fix the firmware update code that was incorrectly ported over from
the idrac puppet module.  This was resulting in the reboot
job and the job queue never being scheduled.

Also, we were incorrectly finishing the job early without checking
all of the statuses

### DIFF
--- a/lib/asm/firmware.rb
+++ b/lib/asm/firmware.rb
@@ -75,6 +75,7 @@ module ASM
       wsman_instance = ASM::WsMan.new(options, :logger => logger)
       firmware_instance.clear_job_queue_retry(wsman_instance)
       force_restart = config["asm::server_update"][cert_name]["force_restart"]
+      logger.debug("Idrac FW update, force restart selected for %s" % cert_name) if force_restart
       pre = []
       main = []
 
@@ -125,7 +126,7 @@ module ASM
 
         # Resets two times
         if attempts < 3
-          transport = ASM::Transport::Racadm.new(ASM::WsMan::Client.endpoint, logger)
+          transport = ASM::Transport::Racadm.new(wsman.client.endpoint, logger)
           transport.reset_idrac
           sleep 60
           retry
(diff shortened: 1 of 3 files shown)


logankimmel:
Marked as DNM until I've finished testing

logankimmel:
@gavin-scott @ripienaar This has been tested and is ready for review

lib/asm/firmware.rb
```diff
@@ -75,6 +75,7 @@ def self.idrac_fw_install_from_uri(config, resource_hash, device_config, logger)
       wsman_instance = ASM::WsMan.new(options, :logger => logger)
       firmware_instance.clear_job_queue_retry(wsman_instance)
       force_restart = config["asm::server_update"][cert_name]["force_restart"]
+      logger.debug("Idrac FW update, force restart: %s" % force_restart)
```
ripienaar: please log certname or similar so that someone seeing these in a log file can get some context of whats going on, similar comment for other debug lines here

lib/asm/firmware.rb
```diff
@@ -143,49 +144,53 @@ def clear_job_queue_retry(wsman)
...
+      logger.debug("First statuses set: %s" % statuses)
+
+      statuses.each do |status|
+        if NO_REBOOT_COMPONENT_IDS.include?(status[:firmware]["component_id"].to_i)
+          status[:desired] = "Completed"
+          status[:reboot_required] = false
         else
-          force_restart ? v[:desired] = "Completed" : v[:desired] = "Scheduled"
-          v[:reboot_required] = true
+          force_restart ? status[:desired] = "Completed" : status[:desired] = "Scheduled"
```
ripienaar: maybe just me but the duplication here seems weird, `status[:desired] = force_restart ? "Completed" : "Scheduled"` seems better, but no biggie just personal taste

lib/asm/firmware.rb
```diff
@@ -277,35 +280,68 @@ def block_until_downloaded(job_id, firmware, wsman)
...
-            lc_job = wsman.get_lc_job(reboot_id)
+            lc_job = wsman.get_lc_job(reboot_job_id)
             reboot_status = lc_job[:job_status]
             logger.debug("Reboot Status: #{reboot_status}")
           end
         end
       end
     end
 
+    def setup_job_queue(config_file, wsman)
```
ripienaar: add some docs and tests for these 2 methods please

lib/asm/firmware.rb
logankimmel: I disagree. I'm keeping it as is :)

logankimmel:
@ripienaar updated

lib/asm/firmware.rb
kaushalgala: hmm, even i like `status[:desired] = force_restart ? "Completed" : "Scheduled"`, but I guess what is already here is also ok.

gavin-scott: You really disagree @logankimmel ? That style is a crime against humanity!

PR #79, opened 2016-11-04
ASM-8191: Adds PXE to specified matching Nic in standard switch of cluster

arunrordell:
Previously we reject pxe network from the vim_nic to add to specific ports

This PR will adds pxe and configure port_group with pxe on configured nic.

### DIFF
--- a/lib/asm/network_configuration.rb
+++ b/lib/asm/network_configuration.rb
@@ -337,26 +337,24 @@ module ASM
     # Returns the information to be used for creating the NIC Team / NIC Bonding
     # Information contains the networks and associated MAC Addresses
     #
+    # @option options [Boolean] :include_pxe flag to include pxe in teams or not
     # @return [Hash] Hash having unique list of networks and corresponding server MAC addresses
     # @example
     #   { [ :TeamInfo => { :networks => [...], :mac_addresses => [ ... ] ] }
-    def teams
+    def teams(opt={})
       @teams ||= begin
         raise("NIC MAC Address information needs to updated to network configuration. Invoke nc.add_nics!") unless @network_config_add_nic
         networks = []
         mac_teams = {}
         partitions = get_all_partitions
-        unless partitions.empty?
-          partitions.each do |partition|
-            network_objects = begin
-              partition.networkObjects.reject { |network| network.type == "PXE" }
-            end.flatten.uniq.compact
+        partitions.each do |partition|
+          network_objects = partition.networkObjects.dup
+          network_objects.reject! { |network| network.type == "PXE" } unless opt[:include_pxe]
 
-            # Need to find partitions which has same set of networks, for team
-            next unless network_objects && !network_objects.empty?
-            network_objects.each do |obj|
-              networks.push(obj).uniq!
-            end
+          # Need to find partitions which has same set of networks, for team
+          next unless network_objects && !network_objects.empty?
+          network_objects.each do |obj|
+            networks.push(obj).uniq!
           end
         end
         @teams = []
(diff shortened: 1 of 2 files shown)


arunrordell:
Reference PR 
https://github.com/dell-asm/asm-deployer/pull/1964

lib/asm/network_configuration.rb
```diff
@@ -348,10 +348,7 @@ def teams
         partitions = get_all_partitions
         unless partitions.empty?
           partitions.each do |partition|
-            network_objects = begin
-              partition.networkObjects.reject { |network| network.type == "PXE" }
-            end.flatten.uniq.compact
-
+            network_objects = partition.networkObjects
```
gavin-scott: I worry this behavior change could break linux LACP configuration or Hyper-V teaming -- neither of those flows expect to find the PXE network in the teams.

It's a bit ugly but the simplest safe fix I can think of is add a parameter like `reject_pxe` as a method arg, have it to default to true and only remove pxe if it is true. The ESXi flow can then pass false as an arg.

lib/asm/network_configuration.rb
```diff
@@ -337,20 +337,26 @@ def reset_virt_mac_addr(endpoint)
...
-              partition.networkObjects.reject { |network| network.type == "PXE" }
-            end.flatten.uniq.compact
+
+            if include_pxe
+              network_objects = partition.networkObjects
+            else
+              network_objects = begin
+                partition.networkObjects.reject { |network| network.type == "PXE" }
+              end.flatten.uniq.compact
+            end
```
gavin-scott: The original code is a little weird. `partition.networkObjects` should be an (already flat, already unique, already non-nil) array of network objects. I think this would work fine to replace the whole thing:

    network_objects = partition.networkObjects.dup
    network_objects.reject! { |network| network.type == "PXE" } unless include_pxe

lib/asm/network_configuration.rb
```diff
@@ -337,20 +337,26 @@ def reset_virt_mac_addr(endpoint)
...
     # @example
     #   { [ :TeamInfo => { :networks => [...], :mac_addresses => [ ... ] ] }
-    def teams
+    def teams(include_pxe=false)
       @teams ||= begin
         raise("NIC MAC Address information needs to updated to network configuration. Invoke nc.add_nics!") unless @network_config_add_nic
         networks = []
         mac_teams = {}
         partitions = get_all_partitions
         unless partitions.empty?
```
gavin-scott: Not part of your changes but can you remove the `unless` here? It's completely unneeded.

lib/asm/network_configuration.rb
```diff
@@ -337,20 +337,26 @@ def reset_virt_mac_addr(endpoint)
     # Returns the information to be used for creating the NIC Team / NIC Bonding
     # Information contains the networks and associated MAC Addresses
     #
+    # params [boolean] flag to include pxe in teams
     # @return [Hash] Hash having unique list of networks and corresponding server MAC addresses
     # @example
     #   { [ :TeamInfo => { :networks => [...], :mac_addresses => [ ... ] ] }
-    def teams
+    def teams(include_pxe=false)
```
gavin-scott: See asm-deployer review, I think it would be nicer to make this an options hash where the only expected key is `include_pxe`. So calling code would look like:

    teams = network_config.teams(:include_pxe => true)

lib/asm/network_configuration.rb
```diff
@@ -337,26 +337,23 @@ def reset_virt_mac_addr(endpoint)
     # Returns the information to be used for creating the NIC Team / NIC Bonding
     # Information contains the networks and associated MAC Addresses
     #
+    # params [Hash] options with include_pxe flag to include pxe in teams or not
     # @return [Hash] Hash having unique list of networks and corresponding server MAC addresses
```
gavin-scott: Need the specific option doc here

gavin-scott: Example: `@option options [Fixnum] :sleep_secs the number of seconds to sleep between inventory runs`

PR #81, opened 2017-01-03
misc: modern ruby and coveralls integration

ripienaar:
Also update rspec to latest and updates the rspec syntax

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ doc
 *.swo
 *.swp
 .idea
+bundle
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,5 @@ sudo: false
 rvm:
   - 1.9.3
   - jruby-19mode
+  - jruby-9.1.5.0
+  - 2.1.9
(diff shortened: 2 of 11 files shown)


gavin-scott:
This is still marked DNM but looks good to go to me. Thanks!

ripienaar:
We should set up a coveralls account that isn't my personal one :) not sure of someone should set it up while logged in as the dell-asm user perhaps. Will see how it works

ripienaar:
seems fine, i squashed and update commit message, pr message etc @gavin-scott

gavin-scott:
Nice!

ripienaar:
@gavin-scott i disabled the thing where coveralls comments on the PRs like above, it can be annoying but also improves visibility, think we should keep it on?

kaushalgala:
IMHO I like the coveralls comment that indicates the coverage and a link to the coveralls report. But again, it is first time experiencing so will feel good :)  If it turns out to be annoying in future, we can turn it off. What do you say @gavin-scott ?

ripienaar:
My personal bad habit of squashing and repeatedly
force pushing a PR while working makes it very annoying. But sure I am all for trailing the setting. Will enable later :)

gavin-scott:
The comments seem good to me? I guess it'll become clear if they're annoying when you do lots of squash and force push? (that's my usual workflow also)

ripienaar:
One thing i didn't mention is if test coverage decrease over a threshold the coveralls check fails just like CI fails 

So the main thing we would care about is being surfaced in another way too. 

Will enable the comments today :)

PR #82, opened 2017-01-19
ASM-8477 Unity puppet

hribab:
Unity REST calls only works with Rest-client 1.6.7

Rest-client 1.8.0 doesn't work because it has issues
with handling redirection with cookies and headers


gavin-scott:
@kaushalgala got this working by monkey-patching rest-client

kaushalgala:
hmm, wondering why we had this PR in the first place :) there was no need to add dependency between dell-asm-util and dell-unity

PR #83, opened 2017-01-25
ASM-8733 Do graceful power offs by default

gavin-scott:
In general it's better to do graceful shutdowns than just shutting off
power. There is no straightforward WS-Man method to implement
"graceful with forced" shutdown functionality -- i.e. issuing a
graceful shutdown and then switching off the power at some point if
the OS doesn't respond. This change makes the default
`wsman.power_off` call execute in this manner.

This specifically addresses some cases that were seen when resetting
ESXi management IPs where that command seems to succeed but after
force powering off the server (the previous functionality of
`wsman.power_off`) and then powering it back on that configuration
change was lost.

### DIFF
--- a/dell-asm-util.gemspec
+++ b/dell-asm-util.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "yard"
   s.add_development_dependency "kramdown"
   s.add_development_dependency "rubocop", "0.37.2"
-  s.add_development_dependency "rspec", "~>2.14.0"
+  s.add_development_dependency "rspec", "~> 3"
   s.add_development_dependency "mocha"
   s.add_development_dependency "puppet"
   s.add_development_dependency "puppetlabs_spec_helper", "0.4.1"
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1071,7 +1071,7 @@ module ASM
     # Update power state of the server.
     #
     # @param options [Hash]
-    # @option options [Symbol|String] :power_state 1 to 10"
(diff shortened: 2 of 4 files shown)


gavin-scott:
@sushilrai this is in general what @kaushalgala and I were looking at for fixing the ESXi teardown issues. I'm going to do some quick tests with this just manually issuing the shutdown calls. If that succeeds can you pick up doing some end-to-end VSAN teardown testing with this and merge it if it looks good? I'll update here when I'm doing testing.

kaushalgala:
Looks good to me

gavin-scott:
Tests are broke, getting "NoMethodError: undefined method `last_comment' for #<Rake::Application:0x00000001ac6c68>" Did rake version just change on us?

gavin-scott:
Had to update to rspec 3 which @ripienaar already did in master to fix tests. Also I was using the wrong ws-man command, needed to do RequestPowerStateChange instead of RequestStateChange. The latter doesn't support graceful shutdown.

gavin-scott:
Some simple end-to-end testing using `wsman_shell.rb` to execute `power_off` and `power_off(:power_off)` (the forced shutdown variant) seems to work. Taking DNM off the title.

sushilrai:
@gavin-scott  as suspected graceful shutdown has other repercussions. One of ESXi OS installation failed with timeout error. Job Queue list shows there is a graceful shutdown job got timed-out.  

image

gavin-scott:
We discussed separately, but `RequestPowerStateChange` doesn't create reboot jobs. Those must have come from another process...

sushilrai:
+1 
Had multiple success with VSAN VDS teardown

PR #84, opened 2017-02-03
ASM-8606 Replaced nokogiri 1.5.10 with 1.6.8 for puppet 3.8

arunrordell:

PR #85, opened 2017-02-04
Revert "ASM-8606 Replaced nokogiri 1.5.10 with 1.6.8 for puppet 3.8"

gavin-scott:
Reverts dell-asm/dell-asm-util#84

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,4 @@ sudo: false
 rvm:
   - 1.9.3
   - jruby-19mode
-  - jruby-9.1.5.0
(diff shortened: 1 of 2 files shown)


gavin-scott:
Test failures are because jruby 9 was added back in and that is currently having issues. I think this is ok to merge as is or tomorrow I can get another branch going that reverts the changes but restores the .travis.yml change to remove jruby 9 from the travis build.

gavin-scott:
Updated with separate commit to remove jruby 9k again to fix the tests.

PR #86, opened 2017-02-06
ASM-8652 Fixes for jruby 9

gavin-scott:
- The timeout command argument was passed as a Fixnum argument to
  Open3.popen3 which fails with "TypeError: no implicit conversion of
  Fixnum into String". Interestingly that fails under ruby 1.9 as
  well, it only seems to work in jruby 1.7

- Pin rainbow to 2.0.x to work around issues with [bundler 0.14 and
  rainbow][1].

[1]: https://github.com/sickill/rainbow/issues/48

### DIFF
--- a/lib/asm/util.rb
+++ b/lib/asm/util.rb
@@ -128,7 +128,7 @@ module ASM
     # Management Network      vSwitch0                     1        0
     # vMotion                 vSwitch1                     1       23
     def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=600)
-      args = [time_out, "env", "VI_PASSWORD=#{endpoint[:password]}", "esxcli"]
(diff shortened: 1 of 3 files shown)


lib/asm/util.rb
```diff
@@ -128,7 +128,7 @@ def self.default_routed_ip
     # Management Network      vSwitch0                     1        0
     # vMotion                 vSwitch1                     1       23
     def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=600)
-      args = [time_out, "env", "VI_PASSWORD=#{endpoint[:password]}", "esxcli"]
+      args = [time_out.to_s, "env", "VI_PASSWORD=#{endpoint[:password]}", "esxcli"]
```
ripienaar: guess it makes sense all things going to a shell would be strings, weird.  PR looks good 👍

PR #87, opened 2017-03-01
ASM-9191 let wsman client options adjust exec behaviour

ripienaar:
Previously WsMan::Client accepted a hash of options but ignored these
options and used hard coded defaults instead

Now it uses these options to influence the defaults thus retaining the
ability to make specific per exec adjustments but allowing for different
defaults that would affect all other methods

### DIFF
--- a/lib/asm/wsman/client.rb
+++ b/lib/asm/wsman/client.rb
@@ -10,9 +10,19 @@ module ASM
 
       def initialize(endpoint, options={})
         missing_params = [:host, :user, :password].reject { |k| endpoint.include?(k) }
+
         raise("Missing required endpoint parameter(s): %s" % [missing_params.join(", ")]) unless missing_params.empty?
+
         @endpoint = endpoint
         @logger = augment_logger(options.delete(:logger) || Logger.new(nil))
+        @default_options = {
+          :selector => nil,
+          :props => {},
+          :input_file => nil,
+          :nth_attempt => 0,
+          :transport_timeout => 300,
+          :retry_on_error => true
+        }.merge(options)
 
         proxy_warn
       end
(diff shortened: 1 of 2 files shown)


ripienaar:
@gavin-scott dont know how you feel about this, but I'd like to reuse this class and not make nagios unhappy, one day maybe even move the logic out of the current check-wsman into this

gavin-scott:
@ripienaar it's not doc'd :-( but you can already pass transport_timeout as a parameter to `WsMan.new` and it will get passed to the `WsMan.client` which will then use that timeout for all ws-man calls: https://github.com/dell-asm/dell-asm-util/blob/fc9b3f8a74fb35b772b678897fbc8a83fc2436dd/lib/asm/wsman/client.rb#L58 Does that meet your needs?

ripienaar:
@gavin-scott ah handy ok, I'll check that out, I tried to fix up some of the yard here but I cant figure out what some of it is intending to say :)

let me see if I can get this working, have a nagios check written already

ripienaar:
@gavin-scott I am not seeing this 

https://github.com/dell-asm/dell-asm-util/blob/master/lib/asm/wsman/client.rb#L11

this takes options, and does nothig with it.

the `exec` method just merges 300 in with whatever is supplied - no taking of client wide defaults.

I could probably make this work rather than the PR here - unless you can see how its supposed to work?

gavin-scott:
You're right. To me it makes sense to take it as a constructor argument, save it and use that value as the default in `exec`.

ripienaar:
@gavin-scott ok, I'll make it do that tomorrow instead of this

ripienaar:
@gavin-scott how does this look now?

gavin-scott:
Looks great, thanks!

PR #88, opened 2017-03-08
ASM-9277 Do not #read a closed file handle, File.read instead

ripienaar:


ripienaar:
I'll look at this

spec/unit/asm/firmware_spec.rb
```diff
@@ -222,7 +222,7 @@
     it "should raise an error when install_from_uri job fails" do
       data = {:name => "xyz", :description => "test file"}.to_json
       data.stubs(:path).returns("/tmp")
-      data.stubs(:read).returns("tmp")
+      File.stubs(:read).with("/tmp").returns("tmp")
```
ripienaar: @logankimmel you do kind of test the unhappy path but stub too much so the problem got hidden :) not really always a good way to stub these file things so it is what it is.

ripienaar:
this coverage decrease is because I added lines but did not add more tests - the line is tested.

ripienaar:
@gavin-scott should be mergable now

PR #89, opened 2017-03-15
ASM-9359 FW Update Debug line throwing "too few args"

logankimmel:
Using StringF notation we need to make sure the argument is a string


logankimmel:
@kaushalgala Can you review?

PR #90, opened 2017-03-23
ASM-9333 Increased time-out for reboot job event from 5 to 15 mnts

arunrordell:
Previously reboot job has 5 minutes to timeout, sometimes resulting
in timeout  error  waiting for LC ready if servers takes more than 5 minutes.

This PR increases the time to 15 minutes to wait for LC to become ready.


kaushalgala:
I know the reason for bumping up the timeout to 15 mins, but it would be good for description documentation on reason to do it

PR #92, opened 2018-01-22
ESTS-139864 support ESXi 6.7 deployment

danielseankim:
Since ESXi 6.7 requires CA cert installation or thumbprint validation,
lib/asm/util.rb file was updated to meet this requirement.

For the purpose of this task, temporary thumbprint validation seemed
as the most appropriate solution, as this method does not require
clean up during the teardown.

A helper method was implemented to grab and parse the thumbprint,
and it is called from the esxcli method to set a command argument.

### DIFF
--- a/lib/asm/util.rb
+++ b/lib/asm/util.rb
@@ -128,10 +128,17 @@ module ASM
     # Management Network      vSwitch0                     1        0
     # vMotion                 vSwitch1                     1       23
     def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=600)
+      unless cmd_array.empty?
+        endpoint[:thumbprint] ||= begin
+          thumbprint_output = esxcli([], endpoint, logger, true, time_out)
+          thumbprint_output.slice(/(?<=thumbprint: )(.*)(?= \(not)/)
+        end
+      end
       args = [time_out.to_s, "env", "VI_PASSWORD=#{endpoint[:password]}", "esxcli"]
       args += ["-s", endpoint[:host],
                "-u", endpoint[:user]
-              ]
+      ]
+      args += ["-d", endpoint[:thumbprint]] if endpoint[:thumbprint]
       args += cmd_array.map(&:to_s)
 
       if logger
@@ -139,10 +146,9 @@ module ASM
         tmp[2] = "VI_PASSWORD=******" # mask password
         logger.debug("Executing esxcli #{tmp.join(' ')}")
       end
-
       result = ASM::Util.run_command_with_args("timeout", *args)
 
-      unless result["exit_status"] == 0
+      if result["exit_status"] != 0 && !cmd_array.empty?
         msg = "Failed to execute esxcli command on host #{endpoint[:host]}"
         logger.error(msg) if logger
         args[2] = "VI_PASSWORD=******" # mask password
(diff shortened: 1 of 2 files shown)


danielseankim:
@gavin-scott Could you review this and merge?

lib/asm/util.rb
```diff
@@ -180,6 +183,24 @@ def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=60
       end
     end
 
+    def self.get_esxi_thumbprint(endpoint, logger=nil, time_out=600)
```
gavin-scott: Most of this is a duplicate from `ASM::Util.esxcli`. Could you just call that method itself if the thumbprint is not provided as an arg? Maybe make it part of the `endpoint` since conceptually I could see we might want to record the thumprint as part of the general endpoint info.

Something like:

    def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=600)
        endpoint[:thumbprint] ||= begin
            thumbprint_output = esxcli(cmd_array, endpoint, logger, true, time_out)
            parse_thumbprint(thumbprint_output)
        end
    end

danielseankim: So I changed the code to something like this:
```
def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=600)
      args = [time_out.to_s, "env", "VI_PASSWORD=#{endpoint[:password]}", "esxcli"]
      args += ["-s", endpoint[:host],
               "-u", endpoint[:user]
      ]

      thumbprint = endpoint[:thumbprint]
      if thumbprint.nil?
        thumbprint_output = ASM::Util.run_command_with_args("timeout", *args)
        unless result["stdout"].include? "thumbprint:"
          msg = "Failed to get esxcli thumbprint on host #{endpoint[:host]}"
          logger.error(msg) if logger
          args[2] = "VI_PASSWORD=******" # mask password
          raise("#{msg}: esxcli #{args.join(' ')}: #{result.inspect}")
        end
        thumbprint = thumbprint_output["stdout"].slice(/(?<=thumbprint: )(.*)(?= \(not)/)
        args += ["-d", thumbprint]
      end

      args += cmd_array.map(&:to_s)
      if logger
        tmp = args.dup
        tmp[2] = "VI_PASSWORD=******" # mask password
        logger.debug("Executing esxcli #{tmp.join(' ')}")
      end

      result = ASM::Util.run_command_with_args("timeout", *args)
      ############### Rest of the method is omitted ###############
```

However, I am not sure if making it backward compatible is worth it. Even though you cannot use the -d option with older esxcli, it will still print out the thumbprint. The only way to make it backward compatible is to add another call to esxcli to get the version info. Let me know what you think.

danielseankim: I also have a PR for asm-automation to use new version of CLI:
https://github.com/dell-asm/asm-automation/pull/607

danielseankim:
@gavin-scott Let me know if latest changes meet your request.

danielseankim:
This PR is dependent on dell-asm/asm-automation#607

lib/asm/util.rb
```diff
@@ -131,9 +131,22 @@ def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=60
...
+      if thumbprint.nil?
+        thumbprint_output = ASM::Util.run_command_with_args("timeout", *args)
+        unless thumbprint_output["stdout"].include? "thumbprint:"
+          msg = "Failed to get esxcli thumbprint on host #{endpoint[:host]}"
+          logger.error(msg) if logger
+          args[2] = "VI_PASSWORD=******" # mask password
+          raise("#{msg}: esxcli #{args.join(' ')}: #{thumbprint_output.inspect}")
+        end
+        thumbprint = thumbprint_output["stdout"].slice(/(?<=thumbprint: )(.*)(?= \(not)/)
+      end
```
gavin-scott: I'd like to spend a little bit working on this with you. This is still duplicated code from the sections below, I was hoping that could all be handled by recursively calling the `esxcli` function. Something like if there is no `cmd_array` specified and no `endpoint[:thumbprint]` I would think we could fall through to the section below and just skip adding the `-d` argument.

danielseankim: Hmm... I thought I saw some issue with recursive call, but we can look at this today.

lib/asm/util.rb
```diff
@@ -152,31 +158,35 @@ def self.esxcli(cmd_array, endpoint, logger=nil, skip_parsing=false, time_out=60
       if skip_parsing
         result["stdout"]
       else
-        lines = result["stdout"].split(/\n/)
-        if lines.size >= 2
-          header_line = lines.shift
-          seps = lines.shift.split
-          headers = []
+        parse_esxcli_result(result["stdout"])
```
gavin-scott: @danielseankim what are these changes about? Did the output format change so we have to change the parsing also?

danielseankim: @gavin-scott the output format changed as we were trying to support recursive call and to refactor the parsing logic.

gavin-scott: Looks like I got confused by the diff, the implementation of the tabular parsing is the same as before. Thanks!

PR #93, opened 2018-01-31
Adding wrapper methods for puppet bolt command executions

sushilrai:
Added wrapper methods for invoking puppet bolt command and scripts. These methods will execute and provide the STDOUT of the command executed on the remote server


gavin-scott:
Outdated

PR #94, opened 2018-01-31
Ests 140275 fix github vulnerabilities scans

pschami:


gavin-scott:
@pschami I'd rather go with https://github.com/dell-asm/dell-asm-util/pull/95 if you can review. That opens up the gemspec so that the consumer (asm-deployer, dell-idrac) can determine the gem versions to use. It also effectively tests this repo against the latest library versions. If we lock down the gemspec here to specific versions then the consumers will start failing immediately until their Gemfiles are updated to match the new versions.

PR #96, opened 2018-03-02
LUD-279 Add support for Mellanox 25Gb NICs

gavin-scott:
Additionally added some specific checks for Intel X710 which was
already supported. Both the X520 and X710 have both 2x10Gb and 4x10Gb
variants but the previous X520 check was specific to 2-port.


gavin-scott:
DO NOT MERGE, still testing...

gavin-scott:
Should be good to go now, removed the DNM in title.

PR #97, opened 2018-03-27
Accept hostname in `ASM::Util.get_preferred_ip`

gavin-scott:
The vmware module uses this method to determine the local IP to use as
an NFS mount. In cases where the ESXi hosts are added to the vcenter
via their hostname rather than IP this was failing.

### DIFF
--- a/lib/asm/util.rb
+++ b/lib/asm/util.rb
@@ -83,22 +83,35 @@ module ASM
       end.ip_address
     end
 
-    # In case of dual NIC appliance with access to non-routable network
-    # method will get the IP address of the appliance which has the access to a particular network
-    # ping the destination IP with specific NIC interface to confirm the access
-    def self.get_preferred_ip(ipaddress)
+    # Return the host IP that routes to the specified host
+    #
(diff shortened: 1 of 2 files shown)


gavin-scott:
@sushilrai this is a miscellaneous issue I ran into yesterday.

sushilrai:
There is a styling error.

lib/asm/util.rb
```diff
@@ -83,30 +83,44 @@ def self.first_host_ip
...
+    #
+    # @return [String] the local host IP address
+    def self.get_preferred_ip(remote_host, logger=nil)
       tries = 0
       max_tries = 10
       preferred = nil
 
+      ip = Resolv.getaddress(remote_host)
+
       while tries < max_tries
```
gavin-scott: @sushilrai any idea why we're doing retry here? I would think that short of adding or removing a network interface that the return value of `ip route get` wouldn't change.

sushilrai: I dont recall this, but possible scenario could be network interface not accessible while resolving the IP address

PR #99, opened 2018-04-11
LUD-522 Deadlock when running "puppet asm process_node"

gavin-scott:
Fixed code to use IO::select which waits until data is available
or EOF on an array of files (stdout and stderr in this case). The
code still reads from those streams line by line in an attempt
to interleave the output to preserve previous behavior.


gavin-scott:
Worked with @mdcondon to get tests added for https://github.com/dell-asm/dell-asm-util/pull/98

Closing that in favor of this.

gavin-scott:
Reviewed with @mdcondon offline, he gives approval.

PR #100, opened 2018-04-30
LUD-421b poll_for_lc_ready to handle API errors

danielseankim:
For the service firmware update task, poll_for_lc_ready breaks often
by the endpoint being in a reboot stage.

Now it handles exception caused by API calls failing to reach iDRAC
servers.

### DIFF
--- a/lib/asm/firmware.rb
+++ b/lib/asm/firmware.rb
@@ -95,7 +95,9 @@ module ASM
       firmware_instance.update_idrac_firmware(main, force_restart, wsman_instance)
 
       # After updating Ensure LC is up and in good state before exiting
-      wsman_instance.poll_for_lc_ready
+      ASM::Util.block_and_retry_until_ready(1800, [ASM::WsMan::RetryException, ASM::WsMan::Error, ASM::WsMan::ResponseError], 60) do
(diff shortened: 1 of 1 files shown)


danielseankim:
@gavin-scott I think you wanted the block_and_retry_until_ready() to go into the method that calls poll_for_lc_ready, but I thought it may cause issues since there are multiple stages that need to be validated in this method.

lib/asm/wsman.rb
```diff
@@ -1261,30 +1261,35 @@ def self.boot_to_network_iso(endpoint, source_address, logger=nil, image_name="m
...
         unless resp[:lcstatus] == "0"
           logger.info("LC status on %s: %s" % [host, Parser.response_string(resp)])
           raise(RetryException)
         end
-        resp
       end
+
+      # Disconnect and detach ISO if necessary
+      resp = get_network_iso_image_connection_info
+      if resp["image_name"]
```
gavin-scott: The behavior changed here where e.g. we don't `detach_iso_image` unless `resp["image_name"]`. Need to restore original behavior.

danielseankim:
@gavin-scott let me know if the update meets your change request.

PR #101, opened 2018-05-14
LUD-959 enable/disable iDRAC front panel LED light

danielseankim:
In addition to putting server and its related components into
maintenance mode, this change is required to set the LED blinker
state through IdentifyChassis method using wsman.

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -21,6 +21,7 @@ module ASM
     SOFTWARE_INSTALLATION_SERVICE = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_SoftwareInstallationService?CreationClassName=DCIM_SoftwareInstallationService&SystemCreationClassName=DCIM_ComputerSystem&SystemName=IDRAC:ID&Name=SoftwareUpdate".freeze
     IDRAC_CARD_ENUMERATION = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/DCIM_iDRACCardEnumeration?InstanceID=iDRAC.Embedded.1#VirtualConsole.1#AttachState".freeze
     APPLY_ATTRIBUTES = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_iDRACCardService?SystemCreationClassName=DCIM_ComputerSystem&CreationClassName=DCIM_iDRACCardService&SystemName=DCIM:ComputerSystem&Name=DCIM:iDRACCardService".freeze
+    SYSTEM_MANAGEMENT_SERVICE = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_SystemManagementService?SystemCreationClassName=DCIM_ComputerSystem&CreationClassName=DCIM_SystemManagementService&SystemName=srv:system&Name=DCIM:SystemManagementService".freeze
 
     # rubocop:enable Metrics/LineLength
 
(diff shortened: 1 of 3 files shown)


lib/asm/wsman.rb
```diff
@@ -1533,5 +1534,31 @@ def identify(timeout)
...
+    # @return [Void]
+    # @raise [ASM::WsMan::ResponseError] when it fails to set the LED state
+    def identify_chassis_off
+      client.invoke("IdentifyChassis",
+                    SYSTEM_MANAGEMENT_SERVICE,
+                    :params => {:identify_state => 0},
+                    :required_params => :identify_state,
+                    :optional_params => :duration_limit,
+                    :return_value => "0")
+    end
```
gavin-scott: Please provide a generic implementation of the `identify_chassis` method that supports all parameters and doc the parameters and their possible values. See this one for example: https://github.com/dell-asm/dell-asm-util/blob/8f32fa0dfd8cd42397a6a2e124c6072c93ceafbe/lib/asm/wsman.rb#L647-L668

Generally we want to keep this as a generic wsman implementation and not necessarily specific to our workflows. If you want to implement `identify_chassis_off` and `identify_chassis_on` as well that just call the new `identify_chassis` method with the appropriate arguments that's fine but probably not necessary.

lib/asm/wsman.rb
```diff
@@ -1533,5 +1534,27 @@ def identify(timeout)
...
+    #
+    # This method implements a call to iDRAC, through WsMan, to change the LED blinker
+    # state. To turn the blinking LED on, identify_state parameter must be passed with
+    # the value of 1. To turn it off, the value must be 0. By default, identy_state is
+    # set to 0.
+    #
+    # @param identify_state [FixNum] number that indicates desired LED blinker state
+    # @return [Void]
+    # @raise [ASM::WsMan::ResponseError] when it fails to set the LED state
+    def identify_chassis(identify_state=0)
```
gavin-scott: This doesn't allow me to set the `:duration_limit` parameter. Sorry for the confusion and delayed response but my original request was to create a generic method that would support all of the parameters of the `IdentifyChassis` method. It should be similar to this:

https://github.com/dell-asm/dell-asm-util/blob/8f32fa0dfd8cd42397a6a2e124c6072c93ceafbe/lib/asm/wsman.rb#L647-L668

lib/asm/wsman.rb
```diff
@@ -1533,5 +1534,27 @@ def identify(timeout)
...
+      if identify_state != 0 && identify_state != 1
```
gavin-scott: You can define `:identify_state` here and get this error checking automatically:

https://github.com/dell-asm/dell-asm-util/blob/8f32fa0dfd8cd42397a6a2e124c6072c93ceafbe/lib/asm/wsman/parser.rb#L147

That also lets you set a nice symbol for it so the caller can say something like `wsman.identify_chassis(:identify_state => :blinking)` rather than `wsman.identify_chassis(:identify_state => 1)`

spec/unit/asm/wsman_spec.rb
```diff
@@ -800,4 +800,36 @@
...
+    it "should turn off the LED using IdentifyChassis method by default" do
+      client.expects(:invoke).with("IdentifyChassis",
+                                   ASM::WsMan::SYSTEM_MANAGEMENT_SERVICE,
+                                   :params => {:identify_state => 0},
+                                   :required_params => :identify_state,
+                                   :optional_params => :duration_limit,
+                                   :return_value => "0").returns("message_arguments:" => "IdentifyState", "message_id" => "SYS001", "return_code" => 0)
+      wsman.identify_chassis
+    end
+  end
```
gavin-scott: If you make the changes requested above probably only one test case is required since it is just pass-through to `client.invoke` at that stage.

lib/asm/wsman.rb
```diff
@@ -1533,5 +1534,28 @@ def identify(timeout)
...
+    # Turn on/off the identifying LED on the iDRAC front panel
+    #
+    # This method implements a call to iDRAC, through WsMan, to change the LED blinker
+    # state. To turn the blinking LED on, identify_state parameter must be passed with
+    # the value of 1. To turn it off, the value must be 0. By default, identy_state is
+    # set to 0.
+    #
+    # @param identify_state [FixNum] number that indicates desired LED blinker state
+    # @param options [Hash]
+    # @option options [String|Fixnum|Symbol] :identify_state option to specify whether identifying LED should be blinking or off
```
gavin-scott: Can you doc the allowable symbol values and their meaning like here: https://github.com/dell-asm/dell-asm-util/blob/8f32fa0dfd8cd42397a6a2e124c6072c93ceafbe/lib/asm/wsman.rb#L659

lib/asm/wsman.rb
```diff
@@ -1533,5 +1534,28 @@ def identify(timeout)
...
+    # @option options [String|Fixnum] :duration_limit option to specify how long the LED should be blinking
```
gavin-scott: Can you doc the duration unit (seconds?)?

PR #103, opened 2018-05-25
LUD-1187b handle timeout issue with LC status check

danielseankim:
After the last commit, with changes that wraps polling for LC
status for the firmware update, it was discovered by QAs that
timeout exceptions still can be caused by clear_job_queue_retry
method. This change wraps the call to WsMan::poll_for_lc_ready in
clear_job_queue_retry method to prevent the issue.


danielseankim:
@gavin-scott can you review and merge this one?

PR #108, opened 2018-06-22
LUD-1453 force to clear iDRAC job queue

danielseankim:
There are few occasions where FW/SW update job fails, due to iDRAC
jobs that are stuck in the queue. To mitigate this type of failures,
we now attempt to force clear the iDRAC job queue if the first
try fails.


danielseankim:
@gavin-scott, @sushilrai please review and merge this PR please.

PR #109, opened 2018-07-05
LUD-1629 increase FW update timeout

danielseankim:


danielseankim:
@gavin-scott please review & merge

PR #110, opened 2018-08-22
VXFM-2413 Add helper method to retrieve partitions by id

gavin-scott:


gavin-scott:
Travis CI failure will be fixed by merging https://github.com/dell-asm/dell-asm-util/pull/111 first

PR #112, opened 2018-08-27
VXFM-2197 Add the support for 4N

sushilrai:
Including method to return list of FQDDs mapped to a specific network. This helps in identifying the NICS mapped to a specific nettwork type

### DIFF
--- a/lib/asm/network_configuration.rb
+++ b/lib/asm/network_configuration.rb
@@ -159,6 +159,18 @@ module ASM
       end.flatten.compact
     end
 
+    # Returns list of FQDDs for the NICS associated with specified network ID
+    #
+    # @params [String] network_id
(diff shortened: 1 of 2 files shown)


lib/asm/network_configuration.rb
```diff
@@ -159,6 +159,14 @@ def macs_for_network(network_id)
       end.flatten.compact
     end
 
+    def fqdds_for_network(network_id)
```
gavin-scott: Can you throw in quick YARD doc and unit tests for this? Should be straight-forward to unit test, there are already tests that have data defined you can use, see e.g. https://github.com/dell-asm/dell-asm-util/pull/110/files

gavin-scott:
Not needed.

PR #119, opened 2019-01-04
VXFM-4193 Updated ruby .travis.yml file for Bundler 2.0

arunrordell:

PR #120, opened 2019-01-09
Revert "VXFM-4193 Updated ruby .travis.yml file for Bundler 2.0"

gavin-scott:
Reverts dell-asm/dell-asm-util#119


gavin-scott:
I believe this is breaking the appliance builds because it requires a version of nokogiri not installed there.

PR #121, opened 2019-01-11
VXFM-4200 Develop the foundation code to enable run Ansible playbook

aimers1975:
Added methods to get vault encrypted values, write ansible playbooks and
inventory files from hash, run ansible, and process an text output
file with ansible result.

### DIFF
--- a/lib/asm/util.rb
+++ b/lib/asm/util.rb
@@ -283,9 +283,9 @@ module ASM
 
     # Run cmd by passing it to the shell and stream stdout and stderr
     # to the specified outfile
-    def self.run_command_streaming(cmd, outfile)
+    def self.run_command_streaming(cmd, outfile, env={})
       File.open(outfile, "a") do |fh|
-        Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
+        Open3.popen3(env, cmd) do |stdin, stdout, stderr, wait_thr|
           stdin.close
 
           files = [stdout, stderr]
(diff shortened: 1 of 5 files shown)


lib/asm/util.rb
```diff
@@ -364,6 +364,106 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+    # the output to the provided output file location.
+    #
+    # @param playbook_file [String] Path to the ansible playbook
+    # @param inventory_file [String] Path to the ansible inventory
+    # @param output_file [String] Path to output file
+    # @param verbose [Boolean] Provide the verbose parameter to ansible run
+    #
+    # @return [void]
+    def self.run_ansible_playbook_with_inventory(playbook_file, inventory_file, output_file, verbose=false)
+      cmd = verbose ? "ansible-playbook -v -i %s %s" % [inventory_file, playbook_file] : "ansible-playbook -i %s %s" % [inventory_file, playbook_file]
```
gavin-scott: Safer not to build a command-line string like this, i.e. above will break if paths have spaces in them. `run_command_streaming` will accept an array of command name, argv[0] and zero or more arguments and not run the command-line through a shell which is a good way to fix this issue. The cmd argument ultimately gets passed to `Process.spawn`: https://ruby-doc.org/core-2.2.0/Process.html#method-c-spawn

lib/asm/util.rb
```diff
@@ -364,6 +364,106 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+    # @param yaml_hash [String] Hash formatted for ansible
+    # @param yaml_file_path [String] Path on where to write ansible yaml
+    #
+    # @return [void]
+    def self.write_ansible_yaml(yaml_hash, yaml_file_path)
+      f = File.new(yaml_file_path, "w")
+      # Ansible has trouble if beginning of vault value is placed on next line
+      yaml_data = yaml_hash.to_yaml.gsub(/: \|\n        !vault/, ":  !vault")
+      f.write(yaml_data)
+      f.close
```
gavin-scott: You're probably fine to just do `File.write(yaml_file_path, yaml_hash.to_yaml.gsub(/: \|\n        !vault/, ":  !vault"))` rather than manually writing and closing the file.

lib/asm/util.rb
```diff
@@ -364,6 +364,106 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+      nil
+    end
+
+    # Use ansible vault to encrypt input string
+    #
+    # @param input_string [String] String to be encrypted using ansible-vault
+    #
+    # @return [String] vault encryption string
+    def self.encrypt_string_with_vault(input_string)
+      cmd = "echo $VAULT_VAL | ansible-vault encrypt_string"
```
gavin-scott: Using an environment variable like this doesn't help with the concern about being able to see the password in the process table (i.e. `ps` output) because it is still running `echo MYSUPERSERCRETPASSWORD` and that will appear in the password table. I.e. the variable gets filled out with the actual value when echo executes it. Also this will fail if the password has special characters with it.

Looking at the ansible-vault docs it looks liek the most secure thing to do would be to "read from stdin" option like `ansible-vault encrypt_string --vault-id dev@./password --stdin-name 'new_user_password'`. That will cause you to have to use something  like `Open3.popen3` so that you can get a handle to the process `stdin` and `write` to it. See e.g. here for a popen3 example: https://github.com/dell-asm/dell-asm-util/blob/a02092828eec45e6eac79cb797adc90d0b8a37fd/lib/asm/util.rb#L274 In your case you would write to stdin before closing it.

lib/asm/util.rb
```diff
@@ -364,6 +364,106 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+      begin
+        result = run_command({"VAULT_VAL" => input_string}, cmd)
+      rescue
+        raise("Error getting vault value: %s" % $!.to_s)
+      end
+      result["stdout"]
```
gavin-scott: Probably want to check the process exit code as well.

lib/asm/util.rb
```diff
@@ -364,6 +364,106 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+      found_recap = false
+      found_end = false
+      result = []
+
+      # TODO: Could parse more details on targets, tasks, and tasks results/logs if we want to
+      File.readlines(ansible_out).each do |line|
+        unless line.valid_encoding?
+          line = line.encode("UTF-8", "binary", :invalid => :replace, :undef => :replace, :replace => "?")
+        end
+        if line =~ /PLAY \[(.*)\]/
```
gavin-scott: Does ansible support writing an output file in a machine-readable format like YAML or JSON or XML? It seems surprising that we'd have to parse what looks like fairly free-form output to be able to understand the results.

aimers1975: yes we need to change /etc/ansible/ansible.cfg to use: stdout_callback = json

lib/asm/util.rb
aimers1975: Changed per feedback

lib/asm/util.rb
aimers1975: Done!

lib/asm/util.rb
aimers1975: Done!

aimers1975:
@gavin-scott I've added tests, but I can't quite tell why the checks are failing.  Can you take a look and see if the Travis CI stuff is due to something I did, or another reason?

gavin-scott:
@aimers1975 / @arunrordell -- tests are failing due to VXFM-4193, I had to revert @arunrordell 's original fix: https://github.com/dell-asm/dell-asm-util/pull/120

@arunrordell any ideas on another way to fix this?

lib/asm/util.rb
```diff
@@ -364,6 +364,110 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+      env = {}
+      env["VAULT"] = options[:vault_password_id] if options[:vault_password_id]
+      env["ANSIBLE_STDOUT_CALLBACK"] = options[:stdout_callback]
+      env["ANSIBLE_HOST_KEY_CHECKING"] = "False" unless options[:host_key_check]
+      args = ["ansible-playbook"]
+      args.push("-v") if options[:verbose]
+      args += ["-i", inventory_file, playbook_file]
+      args.push("--vault-password-file") if options[:vault_password_file]
+      args.push(options[:vault_password_file]) if options[:vault_password_file]
+      cmd = args.join(" ")
```
gavin-scott: Unless I'm mis-reading this has the command injection issue I talked about the other day. You shoudl be able to just pass cmd as an array and the `popen3` will execute it without a shell which will fix the issue.

lib/asm/util.rb
```diff
@@ -364,6 +364,110 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+    # @param input_string [String] String to be encrypted using ansible-vault
+    # @param vault_password_file [String] Path to python script for vault password
+    #
+    # @return [String] vault encryption string
+    def self.encrypt_string_with_vault(vault_password_id, input_string, vault_password_file)
+      raise("Error vault password id required") unless vault_password_id
+      raise("Error vault password file required") unless vault_password_file
+      raise("Error no value to encrypt provided") unless input_string
+      args = ["ansible-vault", "encrypt_string", "--vault-password-file", vault_password_file]
+      cmd = args.join(" ")
```
gavin-scott: Pass `cmd` as an array so that it doesn't execute through the shell.

lib/asm/util.rb
```diff
@@ -364,6 +364,110 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+      env = {"VAULT" => vault_password_id}
+      result = Hashie::Mash.new
+
+      begin
+        Open3.popen3(env, cmd) do |stdin, stdout, stderr, wait_thr|
+          stdin.write(input_string)
```
gavin-scott: Yay!

lib/asm/util.rb
```diff
@@ -364,6 +364,110 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
...
+    #
+    # @param ansible_out [String] Path to ansible output file to parse
+    #
+    # @return [String] Results of run in json format
+    def self.parse_ansible_log(ansible_out)
+      # Check results from output of ansible run
+      out_file = ""
+      File.readlines(ansible_out).each do |line|
+        # Throw out ansible non-json output when a failure occurs.
+        out_file += line unless line =~ /to retry(.*)retry/
```
gavin-scott: This is ok, but is this maybe a result of intermixing stdout and stderr? I.e. do we see these lines because we write stdout and stderr to same file?

aimers1975: Investigated and it looks like the retry line actually is coming from stdout.

aimers1975:
@gavin-scott OK, I think I've addressed your feedback, let me know if you see anything else problematic!

PR #130, opened 2019-03-05
VXFM-4040 Update ansible inventory calls to run with a clean environment

aimers1975:
Environment was causing errors to occur in some cases when ansible
inventory was run outside of a script.


aimers1975:
@gavin-scott OK, sorry about that!

PR #132, opened 2019-03-21
VXFM-5171 Inventory fails as host key checking seems to still occur on Ansible inventories.

aimers1975:
Provide method to remove host file entry which can be called before
ansible run.

### DIFF
--- a/lib/asm/util.rb
+++ b/lib/asm/util.rb
@@ -528,6 +528,22 @@ module ASM
       end
     end
 
+    # Remove host entries in /home/razor/.ssh/known hosts file which can cause ansible run to fail
+    #
+    # @param ips [Array] List of host IPs to search and remove for in known_hosts file
+    # @return [List<String>] List of IPs that could not be removed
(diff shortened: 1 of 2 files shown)


lib/asm/util.rb
```diff
@@ -37,6 +37,7 @@ module Util
     DEVICE_MODULE_PATH = "/etc/puppetlabs/puppet/modules"
     INSTALLER_OPTS_DIR = "/opt/razor-server/tasks/"
     DEVICE_LOG_PATH = "/opt/Dell/ASM/device"
+    HOST_FILE_PATH = "/home/razor/.ssh/known_hosts"
```
gavin-scott: Grab this from the user's environment:

    File.join(ENV["HOME"], ".ssh", "known_hosts")

aimers1975: Will do!

aimers1975:
@gavin-scott Let me know if this is what you had in mind...or any other changes?

aimers1975:
@gavin-scott Hopefully this is good to go!

PR #134, opened 2019-04-05
VXFM-5613 Need to add a timeout to run ansible playbook.

aimers1975:


lib/asm/util.rb
```diff
@@ -409,14 +409,15 @@ def self.run_ansible_playbook_with_inventory(playbook_file, inventory_file, outp
       args.push(options[:vault_password_file]) if options[:vault_password_file]
       result = Hashie::Mash.new
       begin
-        Open3.popen3(env, "/bin/env", *args) do |stdin, stdout, stderr, wait_thr|
-          stdin.close
-          result.stdout      = stdout.read
-          result.stderr      = stderr.read
-          result.pid         = wait_thr[:pid]
-          result.exit_status = wait_thr.value.exitstatus
+        Timeout.timeout(1800) do
```
gavin-scott: Using the ruby `Timeout` is playing with fire (even though we do it in some places). See https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/

Would be better to execute the command itself using the unix `timeout` utility itself like we do here: https://github.com/dell-asm/dell-asm-util/blob/5b1e04073297b5c65d3212ff75c19551794d2860/lib/asm/util.rb#L167

aimers1975:
@gavin-scott This is ready for you to take another look. thanks!

PR #135, opened 2019-04-16
VXFM-5791 - Adding support for Appliance

krp2:


krp2:
@saxens7 @gavin-scott  Can you please review and merge?

PR #136, opened 2019-05-23
VXFM-6182 check number of jobs in the queue after clear request

danielseankim:


danielseankim:
This must be merged before https://github.com/dell-asm/dell-idrac/pull/305

lib/asm/firmware.rb
```diff
@@ -139,14 +141,33 @@ def clear_job_queue_retry(wsman)
         if attempts < 3
           transport = ASM::Transport::Racadm.new(wsman.client.endpoint, logger)
           transport.reset_idrac
-          sleep 60
+          sleep 180
```
danielseankim: @gavin-scott we may need to modify to call wsman.poll_for_lc_ready. There have been some issues where rebooting idrac took longer than 3 minutes. What you think?

danielseankim: Like I said this morning, this may not make a huge difference at this point. Most failures I see are new jobs getting stuck on retry after clearing the queue.

danielseankim:
@gavin-scott can you review/merge this please?

PR #138, opened 2019-05-29
[DNM] VXFM-6275 - Match boot_device pattern for PXE

saxens7:
Match boot device pattern in bios settings for PXE

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1417,7 +1417,11 @@ module ASM
       boot_settings = boot_source_settings
       boot_order_map = {:hdd => "HardDisk.List.1-1", :virtual_cd => "Optical.iDRACVirtual.1-1", :virtual_floppy => "Floppy.iDRACVirtual.1-1"}
       boot_device = Parser.enum_value("BootDevice", boot_order_map, boot_device, :strict => false)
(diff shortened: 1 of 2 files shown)


lib/asm/wsman.rb
```diff
@@ -1417,7 +1417,9 @@ def find_boot_device(boot_device)
       boot_settings = boot_source_settings
       boot_order_map = {:hdd => "HardDisk.List.1-1", :virtual_cd => "Optical.iDRACVirtual.1-1", :virtual_floppy => "Floppy.iDRACVirtual.1-1"}
       boot_device = Parser.enum_value("BootDevice", boot_order_map, boot_device, :strict => false)
-      boot_settings.find { |e| e[:instance_id].include?("#%s#" % boot_device) }
+      instance_id = boot_settings.find { |e| e[:instance_id].include?("#%s#" % boot_device) }
+      #if boot device not found for in boot_settings, look for part of boot_device
+      instance_id ||= boot_settings.find { |e| e[:instance_id].include?("#%s#" % boot_device.split("-").first) }
```
gavin-scott: This seems like it wouldn't work because it would be searching for "#AHCI.Slot.1#" but the boot source settings had "#AHCI.Slot.1-2#"

PR #139, opened 2019-05-30
VXFM-6275 UEFI PXE fails to find BOSS in boot list

saxens7:
Changed find_boot_device to look for pattern


gavin-scott:
@danielseankim the Travis CI failures look like they might be caused by changes from you to `ASM::Firmware`. Can you have a look and fix if needed?

PR #143, opened 2019-08-06
VXFM-6491 Sets ssl verify option for ansible run

arunrordell:


lib/asm/util.rb
```diff
@@ -395,6 +396,7 @@ def self.run_ansible_playbook_with_inventory(playbook_file, inventory_file, outp
                  :timeout => 1800}.merge(options)
       env = {}
       env["VAULT"] = options[:vault_password_id] if options[:vault_password_id]
+      env["CLOUDLINK_VALIDATE_CERTS"] = "False" unless options[:ssl_cloud_link_certs]
```
gavin-scott: Anyway to specify this in the inventory file or the playbook itself? Would be nice not to have to include module-specific code in the generic `run_ansible_playbook_with_inventory` method.

arunrordell: yes @gavin-scott we can specify in the playbook. Closing this PR

arunrordell:
Closing this one as option can be specified in the playbook

PR #144, opened 2019-09-13
VXFM-7352 wsman.first_sata_disk filters by using SATADOM model

danielseankim:


danielseankim:
@gavin-scott can you merge this please?

lib/asm/wsman.rb
```diff
@@ -1778,7 +1778,7 @@ def identify_chassis(options={})
                     :return_value => "0")
     end
 
-    # Get the FQDD of the first SATA Disk present on the node
+    # Get the FQDD of the first SATA Disk (SATADOM model) present on the node
```
gavin-scott: Do you know for sure this method is only used to get the SATADOM? If so then this change seems fine.

danielseankim: @gavin-scott from what I've heard, we are only supporting SATADOM, I believe. However, I can change this to run again with `if bus_protocol == 5` when `sata_disks.empty?` is true.

gavin-scott: I looked through the code and the only spots calling it look like they are looking for satadom:

https://github.com/dell-asm/asm-deployer/blob/59cb2e910455c9490dc85592309aefca95d04fb2/lib/asm/provider/server/server.rb#L565
https://github.com/dell-asm/asm-deployer/blob/59cb2e910455c9490dc85592309aefca95d04fb2/lib/asm/provider/server/server.rb#L635

Might be nice to rename this to `satadom_disk` or something but this seems OK.

PR #155, opened 2019-11-06
 VXFM-8972 High chance linux CO nodes will fail first time and require retry

sushilrai:
Constant Net::SSH::Verifiers::Null is available in the latest Net::SSH version leading to the falure. Replaced this with false which is the valid equivalent of the constant in the current release.

### DIFF
--- a/lib/asm/util.rb
+++ b/lib/asm/util.rb
@@ -338,7 +338,7 @@ module ASM
       Net::SSH.start(server,
                      username,
                      :password => password,
-                     :verify_host_key => Net::SSH::Verifiers::Null.new,
+                     :verify_host_key => false,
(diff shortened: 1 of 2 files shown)


lib/asm/util.rb
```diff
@@ -338,7 +338,7 @@ def self.execute_script_via_ssh(server, username, password, command, arguments=n
       Net::SSH.start(server,
                      username,
                      :password => password,
-                     :verify_host_key => Net::SSH::Verifiers::Null.new,
+                     :verify_host_key => true,
```
gavin-scott: We keep changing this back and forth. AFAIK both MRI and jruby should have the same version of net-ssh and both use the new constant `Net::SSH::Verifiers::Null.new`. Where are you seeing this failure?

PR #156, opened 2019-11-06
 VXFM-8972 High chance linux CO nodes will fail first time and requir…

gavin-scott:
…e retry

Constant Net::SSH::Verifiers::Null is available in the latest Net::SSH version leading to the falure. Replace this with false which is the valid equivalent of the constant in the current release.


gavin-scott:
3.3.0 backport of https://github.com/dell-asm/dell-asm-util/pull/155 which got merged to wrong branch.

PR #158, opened 2019-11-08
VXFM-8464 Make changes to support raid boot options for nodes on a deployment

kannawin:
Checks for H730P and H740P RAID controllers 

Required for PRs:
https://github.com/dell-asm/asm-core/pull/5433
https://github.com/dell-asm/asm-deployer/pull/3145
https://github.com/dell-asm/dell-idrac/pull/315

### DIFF
--- a/lib/asm/wsman.rb
+++ b/lib/asm/wsman.rb
@@ -1821,6 +1821,18 @@ module ASM
       disk_controller[:fqdd]
     end
 
+    # Get the FQDD of the PERC H730P or H740P RAID controller if present on the node
+    #
+    # @return [String] FQDD of the PERC RAID controller if present on the node
+    def perc_controller
+      disk_controllers = controller_views
+      disk_controller = disk_controllers.find { |c| c[:product_name].include?("H730P") }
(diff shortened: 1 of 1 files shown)


lib/asm/wsman.rb
```diff
@@ -1821,6 +1821,18 @@ def boss_controller
       disk_controller[:fqdd]
     end
 
+    # Get the FQDD of the PERC H730P or H740P RAID controller if present on the node
+    #
+    # @return [String] FQDD of the PERC RAID controller if present on the node
+    def perc_controller
+      disk_controllers = controller_views
+      disk_controller = disk_controllers.find { |c| c[:product_name].include?("H730P") }
+      disk_controller ||= disk_controllers.find { |c| c[:product_name].include?("H740P") }
```
gavin-scott: Generally speaking I wouldn't put product-specific logic here in the wsman library code. I would put it higher up the stack on the asm-deployer side.

kannawin: Even with the BOSS logic above it? I was following that logic for this.

PR #160, opened 2019-11-12
VXFM-9100 Add wait for LC ready before clearing job queue

gavin-scott:
Seeing cases where force clearing the job queue failed which appear to
be caused by issuing the command when LC is not ready.


gavin-scott:
Re-targeted to master branch.

